PR #135, opened 2026-03-25
[INFRA] - Added dev/prod Profiles && Volumes (Hot Reload)

### LINKED ISSUE #39: [INFRA] Research Docker profiles for dev and prod environments - ANDREW
Check the Docker profiles system

### LINKED ISSUE #99: [INFRA] - Add nuke backend command to Makefile
# Description
@Nico-Ry Requested a easy command to clear only the backend and any container/volume linked to it! To avoid needing to clear the whole project and rebuild each time during testing!

The goal is thus to add a `make nuke-backend` command that would achieve this!

Mikan95:
# What was done
- **compose file now has two profiles:**
  - dev
  - prod
- **Volumes have now been added to frontend and backend (in dev mode) allowing:**
  - Hot module reload (HMR) for both the frontend and the backend
  - This means any change you do in the code files should be instantly reflected in the container. (No need to refresh or rebuild)
- **nginx now has two conf files:**
  - **UPDATE:** `nginx.conf` has been removed for now as the `prod` build is not functional
  `nginx.conf` (for prod)
  - `nginx.dev.conf` (for dev)
- **Frontend now runs as:** 
  - node container for dev
  - nginx container for prod
- **Make changes:**
  - `make up` now runs the project in **dev** mode
  - `make prod` can be used to run the project in **prod** mode (Does not work yet, needs SSL certificate and proper routing)
  - `make` (on its own) now defaults to `make help` displaying helpful information about make commands
  - `make help` now displays which targets run in **dev** and which run in **prod**. 
(absence of any marking, means it works for both or is irrelevant to the specific target)

## Dev and Prod Profiles
This change enables docker compose to run two different versions of our project. It works by having a shared base setup: `compose.yaml` and two extra files: `compose.dev.yaml` & `compose.prod.yaml` that overwrite certain parts of the services defined in the `compose.yaml`.

### compose.dev.yaml & compose.prod.yaml
These two files also tell compose how to build the image of each service via a `target` keyword. In this case, **dev** and **prod** respectively. You will see this in the `Dockerfile` of each service using the `AS` keyword to define the behaviour to follow when the `target` is **dev** and when it's **prod**. This allows us to cleanly separate development and production logic.

### Frontend Service Example
The clearest example of this is the frontend that has an extra helper stage called `build` in **prod** that runs `npm run build` to compile all the frontend files into static files and then runs as an nginx container to simply serve these static files when requested. This is done as it is much lighter and faster than a full node container.

**Attention** do NOT confuse this frontend nginx with the entrypoint nginx. The prior acts as a simple file delivery service whereas the latter acts as the security guard to our whole app!

### Port Changes Dev vs Prod
| Build | Port | Access Point |
| ------ | ------ | ----- |
| **dev** | 1024 (no change) | `http://localhost:1024` |
| **prod** | 8080 or 8443 | `http://localhost:8080` redirects to `https://localhost:8443` |

## Volumes Added in Dev Mode
Volumes have been added in dev mode for both the frontend and the backend, this means you should now be able to use hot module reload when running the full stack and no longer need to run the front/back locally.

This also means that for most changes you will not need to rebuild the whole stack or your whole container to have changes take place. This fixes #99 requested by @Nico-Ry.

**Note** however that changes to prisma or the database structure will still require a full rebuild to be reflected. Hopefully, though this will be less frequent than for every single change that occurs in the nestJS code files.  
Depending on what you are trying to achieve it may also sometimes be necessary to run a `make down-v` to delete volumes (persistant data).

_If this becomes heavy or time consuming, get in touch and I'll look at adding a command that nukes and rebuilds only the backend and database containers instead of the whole stack!_

## Nginx Dual Config

**UPDATE:** `nginx.conf` has been removed for now as the `prod` build is not functional

Following the changes to the frontend container in **dev** vs **prod** the nginx container now needs two different configs to handle the frontend as a :
1. **node** (React + Vite) container using port 5173 (Vite default)
2. **nginx** container using port 80 (nginx default)

### Port Confusion
You may be confused about the ports here and what they represent, for us as developers or clients in prod they don't change anything really. When we talk about ports here we are talking about the ports inside the containers, this is only relevant for the docker network and how the nginx container gets in touch with the frontend vs the backend.

### What This Means

```
=====================================================
                 DEVELOPMENT (DEV)
=====================================================

                Docker Network (internal)

        ┌───────────────┐
        │     nginx     │
        │   container   │
        │               │
        │ proxies to →  │
        └───────┬───────┘
                │
                │ http://frontend:5173
                │
        ┌───────▼─────────┐
        │   frontend      │
        │ React + Vite    │
        │ Node dev server │
        │                 │
        │ listens on 5173 │
        └─────────────────┘


✔ Vite dev server runs inside Node
✔ Hot reload enabled
✔ nginx talks to port 5173



=====================================================
                 PRODUCTION (PROD)
=====================================================

                Docker Network (internal)

        ┌───────────────┐
        │     nginx     │
        │   container   │
        │               │
        │ proxies to →  │
        └───────┬───────┘
                │
                │ http://frontend:80
                │
        ┌───────▼─────────┐
        │   frontend      │
        │ nginx static    │
        │ file server     │
        │                 │
        │ listens on 80   │
        └─────────────────┘


✔ Built static files served by nginx
✔ No Node runtime
✔ nginx talks to port 80

```
## Makefile Changes
The Makefile has been quite heavily reorganised and certain parts like the `help` target has now been moved to its own .mk file.

How commands are organised when displayed through `make help` has also seen an overhaul.

Overall the Makefile still functions pretty identical to what it use to with the difference of adjusting commands to run the correct profiles -> **dev** vs **prod**.

Ansi colour codes have also been changed to facilitate their usage:
previously they were between double quotes, now they aren't anymore.
```
// Before
GREEN		= "\033[92m"

// After
GREEN		= \033[92m
``` 
This means when previously a echo call from the Makefile may have looked like this: 
```
@echo $(CYAN)"<Removing Stopped Containers>"$(RES)  <--- colour codes outside of double quotes
```
They now must be contained within the double quotes:
```
@echo "$(CYAN)<Removing Stopped Containers>$(RES)"
```

### Frontend Hot Reload
1. Start containers in **dev** mode:
```
make up
```
2. Access the app through localhost:
```
http://localhost:1024
```
3. Edit the home.tsx file in any way you like:
```
Try changing some text or the colour of a button for example!
```
4. See changes happen near instantaneously

### Backend Hot Reload
Counting on you here @Nico-Ry ^^'

### make commands
**Make prod should not work yet I will add proper SSL certificate and other things later on.**

| target | expected result |
| ------ | ------- |
| help | shows list of commands with dev and prod tags |
| up | starts containers in dev mode |
| nuke | should clear all volumes (Even frontend and backend volumes created after `make up`) |
| _all other targets_ | should still work as intended |

closes #39


proxy/nginx.conf
```diff
@@ -1,30 +1,22 @@
-# required by nginx, left empty for now
```
Nico-Ry: I would recommend not to upload this yet if prod doesn't actually works, it should be a separate PR or something like it

Nico-Ry (changes requested):
everything else looks good, I'll wait for changes to test the PR

Nico-Ry (approved):
# Works,  tested  : 
* make option works
* live changes to files  are applied to the website
---
## Note
* ` make prod`  will not work after a `make nuke`  , but that was implied at the PR level
* only changes in Prisma and dependencies need a make nuke to apply

PR #157, opened 2026-04-03
[Makefile] - Reworked dev setup, nuke cleanup, and guard `make up`

### LINKED ISSUE #136: [MAKEFILE] - Update 'make setup'
# Description
make setup currently does not handle all needed elements, notably:
- DATABASE_URL
- JWT_SECRET

The goal would be to update it for this functionality as well as potentially add a couple more quality of life improvements.

Mikan95:
## What was done
- Rewrote `setup.mk` from scratch — declarative config, smarter check-before-prompt flow
- Added `DEV_ONLY_ENV_VARS` support so dev-only secrets stay out of `.env.example`
- `make up` now runs `setup` as a prerequisite — stack won't start with missing files
- `make nuke` now also deletes `.env` and all secret files, completing the full wipe
- `nuke` warning message updated to explicitly list what will be deleted
- Removed sensitive env vars from .env.example and compose.yaml
- Added sensitve (dev-only) env vars to compose.dev.yaml

closes #136 

## Why

Getting the stack running for the first time required too many manual steps — copying `.env.example`, filling in placeholders, creating secret files by hand. The goal was a zero-friction dev startup: clone, run `make up`, and the stack starts. Hardcoded defaults handle everything automatically, and the setup is modular enough that adding a new secret or env var is a one-liner. `make nuke` was also leaving stale config behind, which made it unreliable as a true reset tool.

## `setup.mk` rewrite

The old setup always recreated files and had a separate target per secret, making it brittle to extend. The new version defines required files, `.env` defaults, dev-only env vars, and secrets as **declarative lists** at the top of the file — adding new items requires only a one-line change to the relevant list.

`make setup` now **checks first**: it iterates over `REQUIRED_FILES` and lists anything missing. If all files are present, it exits silently. Only if something is missing does it ask whether to apply the default setup.

When the user confirms, `_setup-apply` runs — this **always recreates all required files from scratch**, regardless of which ones were actually missing. This is intentional: it guarantees a consistent, known-good state. `.env` is rebuilt from `.env.example` (comments stripped), dev-only vars are prepended, defaults are applied, and every secret file is overwritten with its default value.

## `make nuke` completes the full wipe

Previously `nuke` removed containers, images, and volumes, but left `.env` and secret files on disk. This meant re-running `make up` after a nuke would silently reuse stale config. Now `nuke` deletes everything listed in `REQUIRED_FILES`, so the state of the repo after a nuke is truly clean and `make up` will prompt for setup from scratch.

## How to test

**`make setup` / `make up`**
- With all required files present: `make up` should start the stack without prompting
- Delete one required file (e.g. `rm secrets/backend_pw.txt`) and run `make up` — it should list the missing file and ask if you want the default setup applied
- Answer `n` — it should abort with a message
- Answer `y` — it should recreate all missing files and then start the stack
- Run `make setup` directly when all files are present — should print `✓ All required files present` and exit

**`make nuke`**
- Run `make nuke` and confirm with `y`
- Verify containers, volumes, and images are removed (`make ps`, `docker image ls`)
- Verify `.env` and all files under `secrets/` are gone
- Run `make up` — it should detect missing files and prompt for default setup

## How to add required items to `make setup`

You can consult [this](https://github.com/team-Segfault-Society-42/Ft_transcendence/wiki/Environment-Variables-&-Secrets-%E2%80%94-Where-Does-It-Go%3F) wiki page for guidelines on how to achieve this.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -4,4 +4,3 @@ POSTGRES_DB=your_database_name#ex: postgresql
 POSTGRES_USER=your_db_superuser_name#ex: postgresql_superuser
 BACKEND_USER=your_backend_user_name#ex: backend_user
 DATABASE_URL=postgresql://$BACKEND_USER:$BACKEND_PW@db:5432/$POSTGRES_DB?schema=public#ex:DATABASE_URL=postgresql://backend_user:(password-from-backend_pw.txt)@db:5432/postgresql?schema=public
-JWT_SECRET=change-this-to-a-long-random-secret
--- a/.env.example
+++ b/.env.example
@@ -3,4 +3,3 @@ POSTGRES_VERSION=postgres_image_tag#ex: 18-alpine
 POSTGRES_DB=your_database_name#ex: postgresql
 POSTGRES_USER=your_db_superuser_name#ex: postgresql_superuser
 BACKEND_USER=your_backend_user_name#ex: backend_user
-DATABASE_URL=postgresql://$BACKEND_USER:$BACKEND_PW@db:5432/$POSTGRES_DB?schema=public#ex:DATABASE_URL=postgresql://backend_user:(password-from-backend_pw.txt)@db:5432/postgresql?schema=public
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ SHOW_LOGS			= docker compose logs
 # ══════════════════════════════════════════════════════
 ##@ START STACK
 
-up: ##  Build and run all containers [DEV]
+up: setup ##  Build and run all containers [DEV]
 	@echo "$(GOLD)Building in Dev Mode$(RES)"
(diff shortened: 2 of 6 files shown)


simo1616 (approved):
Tested locally on `infra/better_make_up` and everything worked really well.

Validated:

* `make setup` detects missing required files and recreates `.env` + secret files correctly after confirmation
* `make up` correctly triggers setup when required files are missing
* `make nuke` now removes containers, volumes, `.env`, and all secret files
* after `make nuke`, `make up` prompts again and rebuilds the required setup correctly
* `DATABASE_URL` and `JWT_SECRET` are now provided through dev setup and exposed only in `compose.dev.yaml`
* `compose.yaml` no longer contains those dev-only variables

Honestly, thank you for making our lives easier with this setup rework — this is really clean and super helpful for the team.

Nice Job well done

PR #164, opened 2026-04-10
[BACKEND][AUTH] Apply JwtAuthGuard globally with @Public() for auth routes

### LINKED ISSUE #163: [BACKEND][AUTH] Apply JwtAuthGuard globally with @Public() for auth entry routes
## Scope

Move backend authentication from manual per-route protection with `@UseGuards(JwtAuthGuard)` to a global-guard model where routes are protected by default and only explicitly marked public routes bypass authentication.

This includes introducing a reusable `@Public()` decorator, updating `JwtAuthGuard` to read route metadata, and registering the guard globally so the behavior applies across the backend.

## Tasks

* [x] create `@Public()` decorator in auth module
* [x] define reusable `IS_PUBLIC_KEY` metadata key
* [x] update `JwtAuthGuard` to read public metadata through `Reflector`
* [x] keep JWT validation flow for protected routes
* [x] preserve cookie-first token extraction with Authorization header fallback
* [x] register `JwtAuthGuard` as a global guard
* [x] mark `POST /auth/register` as public
* [x] mark `POST /auth/login` as public
* [x] mark `POST /auth/logout` as public
* [x] remove route-level `@UseGuards(JwtAuthGuard)` from `/auth/me`
* [x] verify `/auth/me` fails without auth
* [x] verify `/auth/me` works with valid cookie
* [x] verify public auth routes still work without token
* [x] verify non-public routes are protected by default

## Why

The backend currently has JWT auth working, but route protection was based on manually adding `@UseGuards(...)` to each protected endpoint.

That approach has 3 problems:

* it is easy to forget protecting a new route
* it repeats auth protection logic across controllers
* it is less secure because public access is the default unless a decorator is added

This issue changes the backend to a safer model:

* protected by default
* public only when explicitly marked with `@Public()`

That makes the auth layer cleaner, more scalable, and harder to misuse.

## Current state

Already implemented before this issue:

* bcrypt password hashing
* JWT issuing
* HttpOnly cookie-based auth
* logout flow
* dedicated `JwtAuthGuard`
* `/auth/me` protected
* Swagger integration through nginx `/api`

This issue focuses only on the global auth-protection step and does not include OAuth or 2FA.

Nico-Ry:
## Summary

This PR closes **#163** by moving authentication from manual route protection to a **global guard model with explicit public routes**.

---

## Key change

```txt
Before → public by default
After  → protected by default
```

---

### Global Guard

* `JwtAuthGuard` registered using `APP_GUARD`
* applies to ALL routes

### Public decorator

* `@Public()` added
* used for:

  * register
  * login
  * logout

### Controller cleanup

* removed `@UseGuards`
* `/auth/me` now protected automatically

---

### Before

```bash
curl http://localhost:1024/api/users
```

✅ Returned:

```json
[...users...]
```

---

### After

```bash
curl -i http://localhost:1024/api/users
```

❌ Returns:

```json
401 Unauthorized
```

---

### 1. Protected route (no auth)

```bash
curl -i http://localhost:1024/api/users
```

Expected:

```json
401 Unauthorized
```

---

### 2. Register (public)

```bash
curl -i -X POST http://localhost:1024/api/auth/register \
  -d '{"email":"test@example.com","password":"secret12","username":"testuser"}'
```

Expected:

```json
201 Created + user data
```

---

### 3. Login (store cookie)

```bash
curl -i -c cookies.txt -X POST http://localhost:1024/api/auth/login \
  -d '{"email":"test@example.com","password":"secret12"}'
```

Expected:

```txt
Set-Cookie: access_token=...
```

---

### 4. Access protected route WITHOUT cookie

```bash
curl -i http://localhost:1024/api/auth/me
```

Expected:

```json
401 Unauthorized
```

---

### 5. Access WITH cookie

```bash
curl -i -b cookies.txt http://localhost:1024/api/auth/me
```

Expected:

```json
200 OK + user data
```

---

### 6. Access `/users` with auth

```bash
curl -i -b cookies.txt http://localhost:1024/api/users
```

Expected:

```json
200 OK + users list
```

---

### 7. Logout

```bash
curl -i -b cookies.txt -c cookies.txt -X POST http://localhost:1024/api/auth/logout
```

Expected:

```txt
cookie cleared
```

---

### 8. Test again

```bash
curl -i -b cookies.txt http://localhost:1024/api/users
```

Expected:

```json
401 Unauthorized
```

---

### Cookie-based authentication (HTTP)

The guard now enforces a **cookie-based authentication model** for HTTP requests:

* JWT is read from `access_token` HttpOnly cookie
* removed `Authorization: Bearer` fallback for HTTP to keep a **single consistent auth flow**

This aligns with the current frontend behavior and simplifies the security model.

---

### WebSocket support

The guard preserves compatibility with WebSocket authentication:

* token is extracted from the handshake cookies
* authenticated user is attached to `client.data.user`

---

### Secure cookie flag

```ts
secure: false
```

* kept for local development (HTTP)
* will be moved to an environment-based configuration when HTTPS is enabled in production

---

### Code cleanup

* removed debug logging from JWT guard
* removed obsolete comments in `users.controller.ts`
* improved type safety with explicit `JwtPayload`, `AuthRequest`, and `AuthSocket`

---

## Result

* secure-by-default backend
* explicit public routes
* cleaner and more maintainable auth logic
* consistent cookie-based authentication model
* ready for future extensions (OAuth, WS auth)

---

closes #163

### DIFF
--- a/backend/src/auth/auth.controller.ts
+++ b/backend/src/auth/auth.controller.ts
@@ -1,86 +1,89 @@
-import { Body, Controller, Post, Get, UseGuards, Req, Res } from '@nestjs/common';
+import { Body, Controller, Post, Get, Req, Res } from '@nestjs/common';
 import type { Response } from 'express';
+import { ApiTags, ApiOperation, ApiResponse, ApiBody } from '@nestjs/swagger';
 import { AuthService } from './auth.service';
 import { RegisterDto } from './dto/register.dto';
 import { LoginDto } from './dto/login.dto';
-import { JwtAuthGuard } from './jwt-auth.guard';
-import { ApiTags, ApiOperation, ApiResponse, ApiBody } from '@nestjs/swagger';
+import { Public } from './public.decorator';
+import type { AuthRequest } from './jwt-auth.guard';
 
 @ApiTags('Auth')
 @Controller('auth')
 export class AuthController {
-		constructor(private readonly authService: AuthService) {}
+	constructor(private readonly authService: AuthService) {}
 
-		@ApiOperation({ summary: 'Register a new user' })
-		@ApiBody({ type: RegisterDto })
-		@ApiResponse({
-				status: 201,
-				description: 'User registered successfully',
-		})
-		@ApiResponse({
-				status: 400,
-				description: 'Invalid input data',
-		})
-		@Post('register')
-		register(@Body() registerDto: RegisterDto) {
-				return this.authService.register(registerDto);
-		}
+	@Public()
+	@ApiOperation({ summary: 'Register a new user' })
+	@ApiBody({ type: RegisterDto })
+	@ApiResponse({
+		status: 201,
+		description: 'User registered successfully',
+	})
+	@ApiResponse({
+		status: 400,
+		description: 'Invalid input data',
+	})
+	@Post('register')
+	register(@Body() registerDto: RegisterDto) {
+		return this.authService.register(registerDto);
+	}
 
-		@ApiOperation({ summary: 'Log in a user and set an HttpOnly JWT cookie' })
-		@ApiBody({ type: LoginDto })
-		@ApiResponse({
-				status: 201,
-				description: 'Login successful',
-		})
-		@ApiResponse({
-				status: 401,
-				description: 'Invalid credentials',
-		})
-		@Post('login')
-		async login(
-				@Body() loginDto: LoginDto,
-				@Res({ passthrough: true }) res: Response,
-		) {
-				const { access_token } = await this.authService.login(loginDto);
+	@Public()
+	@ApiOperation({ summary: 'Log in a user and set an HttpOnly JWT cookie' })
+	@ApiBody({ type: LoginDto })
+	@ApiResponse({
+		status: 201,
+		description: 'Login successful',
+	})
+	@ApiResponse({
+		status: 401,
+		description: 'Invalid credentials',
+	})
+	@Post('login')
+	async login(
+		@Body() loginDto: LoginDto,
+		@Res({ passthrough: true }) res: Response,
+	) {
+		const { access_token } = await this.authService.login(loginDto);
 
-				res.cookie('access_token', access_token, {
-						httpOnly: true,
-						secure: false,
-						sameSite: 'lax',
-						maxAge: 60 * 60 * 1000,
-				});
+		res.cookie('access_token', access_token, {
+			httpOnly: true,
+			secure: false,
+			sameSite: 'lax',
+			maxAge: 60 * 60 * 1000,
+		});
 
-				return { message: 'Login successful' };
-		}
+		return { message: 'Login successful' };
+	}
 
-		@ApiOperation({ summary: 'Log out the current user and clear the auth cookie' })
-		@ApiResponse({
-				status: 201,
-				description: 'Logout successful',
-		})
-		@Post('logout')
-		logout(@Res({ passthrough: true }) res: Response) {
-				res.clearCookie('access_token', {
-						httpOnly: true,
-						secure: false,
-						sameSite: 'lax',
-				});
+	@Public()
+	@ApiOperation({ summary: 'Log out the current user and clear the auth cookie' })
+	@ApiResponse({
+		status: 201,
+		description: 'Logout successful',
+	})
+	@Post('logout')
+	logout(@Res({ passthrough: true }) res: Response) {
+		res.clearCookie('access_token', {
+			httpOnly: true,
+			secure: false,
+			sameSite: 'lax',
+		});
 
-				return { message: 'Logout successful' };
-		}
+		return { message: 'Logout successful' };
+	}
 
-		@ApiOperation({ summary: 'Get the currently authenticated user' })
-		@ApiResponse({
-				status: 200,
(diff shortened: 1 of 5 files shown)


backend/src/auth/jwt-auth.guard.ts
Mikan95: Noticed that `extractTokenFromRequest` falls back to checking the `Authorization: Bearer` header if no cookie is found. 
From my understanding this is mainly used for mobile or service-to-service clients, so I'm not sure this fallback is needed?

The cookie-only path stricter/safer — `httpOnly: true` means the token is invisible to JavaScript and can't be extracted or reused outside the browser session.

The Bearer fallback means a raw token string is enough to authenticate, which slightly widens the attack surface without adding anything for the frontend.

Might be worth removing `extractTokenFromHeader` and its call site unless there's a reason for it I'm missing?

backend/src/auth/auth.controller.ts
Mikan95, line 51: I know this has been done to function with the dev build (HTTP only)! No problem with that, but we will need to update this to `true` (as well as any other similar thing) for the production build!

Just drawing attention to it! No need to do anything here 👍

Mikan95 (commented):
Everything functions great as far as I can tell. Testing instructions were clear and easy to follow!

Please fix the merge conflicts, and take a look at my comments below and I'll gladly approve the PR! :)

backend/src/auth/auth.controller.ts
Nico-Ry: Yess, this is intentionally `secure: false ` for the current local HTTP dev setup, otherwise the auth cookie would not be sent at all on localhost. I’ll keep it as-is for this PR and switch it to an env-based production config once HTTPS / prod profile is in place.
What I would do later is move toward something like:
```ts
secure: process.env.NODE_ENV === 'production',
```
or a dedicated env flag.

backend/src/auth/jwt-auth.guard.ts
Nico-Ry: Good catch. The Bearer fallback was originally kept as a transition/testing path, but for the current browser-oriented cookie auth flow it is not really needed anymore. I’m going to remove the HTTP header fallback here so the auth model stays consistent and cookie-only.

Mikan95: Great! Thanks :)

backend/src/auth/auth.controller.ts
Mikan95: Perfect! Sounds good! I'll get to work on the `NODE_ENV` variable asap! :+1:

backend/src/auth/jwt-auth.guard.ts
```diff
@@ -46,63 +58,42 @@ export class JwtAuthGuard implements CanActivate {
...
+				client.data.user = payload;
 			}
-			else if (context.getType() === 'ws') {
-
-				context.switchToWs().getClient().data.user = payload;
 
-			}
 			return true;
 		} catch {
 			throw new UnauthorizedException('Invalid or expired token');
```
nasdhn, line 75: Based on our tests, with NestJS WebSockets, if you throw a standard UnauthorizedException, the server doesn't necessarily close the connection, and the message can still reach its destination in some cases. By adding `if (contextType === 'ws') return false;` when the token is missing or invalid, you ensure that NestJS ignores the direct message. This solution is more robust and prevents access to our service logic (such as `join_game`) by an unlogged-in user.

in the catch before throw : 
`if (contextType === 'ws') return false;`

backend/src/auth/jwt-auth.guard.ts
nasdhn, line 56: Same here, if the token is missing its better to stop any event

Before throw : 
`if (contextType === 'ws') return false;`

nasdhn (commented):
In the Guard, you need to add `if (contextType === 'ws') return false;` before each throw.

With WebSockets, an `UnauthorizedException` isn't always enough to stop execution: `return false` is the only way to guarantee that NestJS ignores the message and doesn't trigger our services (e.g., `join_game`) without a token. It used to work without it because my code was more basic (without a reflector), but with your new, cleaner structure, `return false` has become mandatory for the blocking to be effective.

backend/src/auth/jwt-auth.guard.ts
Nico-Ry: Good catch, thanks for pointing that out.

I was only testing the HTTP flow, so I didn’t notice the difference in behavior for WebSockets. Your comment makes sense — returning `false` is safer in the WS context to guarantee the event is fully blocked before reaching the gateway.

I’ll update the guard to:

* return `false` for WS when the token is missing or invalid
* keep throwing `UnauthorizedException` for HTTP

That keeps the behavior consistent and ensures the guard properly enforces the zero-trust model for real-time events as well.

backend/src/auth/jwt-auth.guard.ts
Nico-Ry: same as the other. Thanks!

nasdhn (approved):
The changes have been made, in WebSockets, an exception doesn't always stop the execution. Returning false is the only way to guarantee that the server ignores the event and doesn't trigger our services (like join_game) without a token. It worked without it before because my old code was simpler, but with this new structure, it’s required to actually block the access. Everything is good, good job!

Issue #172, opened 2026-04-15
[BUG] - Frontend Hardcoded Paths Break Production Build


Mikan95:
## Description

Pages and services in the frontend hardcode `http://localhost:1024` as the server
address. This breaks in production where the app runs on a different domain over HTTPS.

## Goal

- Replace hardcoded URLs with relative paths so the frontend works in both dev and
production without any configuration changes.
- Route `Home.tsx` through axios in `userService.ts` instead of `fetch`

## Scope

- **Frontend**

## Proposed Tasks

- [ ] `userService.ts` — replace `baseURL: "http://localhost:1024/api/"` with
  `baseURL: "/api/"`
- [ ] `Home.tsx` — replace `fetch("http://localhost:1024/api/game/create")` with
  `fetch("/api/game/create")` (or route through the existing axios instance in
  `userService` for consistency)
- [ ] `Game.tsx` — replace `io("http://localhost:1024", { path: "/socket.io/" })`
  with `io({ path: "/socket.io/" })` — omitting the URL tells Socket.IO to connect
  to the same origin the page was loaded from

## Notes

The browser automatically resolves relative paths against the current origin, so the
same code works on `localhost:1024` in dev and `https://your-domain:8443` in prod.

Note that `Home.tsx` uses raw `fetch` while the rest of the app uses axios via
`userService`. The fix is a good opportunity to align it with the rest of the
codebase, but that can be a separate decision.

## Using Axios for `/api/game/create`
Right now Home.tsx makes its network call like this:

```
fetch("http://localhost:1024/api/game/create", { method: "POST" })
```

> fetch is the browser's built-in way to make HTTP requests, where we get back a raw response and have to handle everything ourselves.

For consistency and cookie handling it may be worth changing this to use axios like the rest of the app 
```typescript
//userService.ts
const api = axios.create({
    baseURL: "/api/",
    withCredentials: true,   // ← automatically sends auth cookie
})
```
> The axios instance in userService has withCredentials: true baked in, so every call through it automatically includes the cookie.

fetch **does not** send cookies by default, unless explicitly asked for. So considering we have just made `/api/game/create` a protected route, the create game call would return `401 Unauthorized` for any logged-in user. (silently, with just a console log).

Mikan95:
This has been closed by #174 Pull Request.

PR #174, opened 2026-04-16
fix(prod): align frontend requests, websocket origins, and VPS deploy

simo1616:
## Summary

This PR aligns the app with the new production setup introduced by the production nginx/proxy work.

## What was done

- replaced hardcoded frontend API URLs with same-origin relative paths
- updated Socket.IO/frontend app flow to work with the current origin in production
- added backend WebSocket CORS origins through environment-based config
- added `CORS_ORIGINS` in `compose.dev.yaml`
- added `CORS_ORIGINS` in `compose.prod.yaml`
- updated the DigitalOcean deploy workflow to use `make prod` instead of the dev flow

## Why

After the production nginx/proxy PR was merged, the app was reachable in prod, but some features still used hardcoded `localhost:1024` paths and dev-only assumptions.

This PR aligns the frontend/backend app layer with the new prod setup so requests now go through the correct origin in both dev and prod.

## Testing

- [x] `make prod` builds successfully
- [x] HTTPS endpoint responds correctly on `https://127.0.0.1:8443`
- [x] frontend loads in production mode
- [x] requests now go to `https://127.0.0.1:8443/api/...` instead of `http://localhost:1024/...`
- [x] verified nginx HTTP -> HTTPS redirect
- [x] updated VPS deploy workflow to run the production stack

### Test details

Performed local production testing with:

- `make setup`
- updated `.env` with a valid IP in `DOMAIN`
  - `127.0.0.1` for same-machine local testing
  - or the host machine IP (for example from `hostname -I`) for LAN testing
- `make prod`

Verified:
- proxy exposes `8080` and `8443`
- `curl -I http://127.0.0.1:8080` returns a 301 redirect
- `curl -vk https://127.0.0.1:8443` returns `200 OK`
- browser loads the frontend over HTTPS
- frontend no longer calls hardcoded `localhost:1024` endpoints

Note: some remaining errors are now app/auth logic related (`401 Unauthorized`) rather than prod-origin / hardcoded-path issues.

## Checklist

- [x] Target branch is `dev`
- [x] No `.env` or secrets committed
- [x] Changes remain focused on prod alignment / deployment

### DIFF
--- a/backend/src/modules/game/game.gateway.ts
+++ b/backend/src/modules/game/game.gateway.ts
@@ -87,7 +87,7 @@ export class GameGateway implements OnGatewayConnection, OnGatewayDisconnect {
   @SubscribeMessage('play_move')
   async handlePlayMove(
     @MessageBody() body: PlayMoveDto,
-    @ConnectedSocket() client: Socket,
+    @ConnectedSocket() client: AuthSocket,
   ) {
     try {
       const newGameState = await this.gameService.playMove(
@@ -108,7 +108,7 @@ export class GameGateway implements OnGatewayConnection, OnGatewayDisconnect {
   @SubscribeMessage('request_replay')
   handleRequestReplay(
     @MessageBody() body: { gameId: string },
-    @ConnectedSocket() client: Socket,
+    @ConnectedSocket() client: AuthSocket,
   ) {
     try {
       const updateGame = this.gameService.requestReplay(body.gameId, client.id);
--- a/backend/src/modules/game/game.gateway.ts
+++ b/backend/src/modules/game/game.gateway.ts
@@ -16,9 +16,20 @@ import { UseGuards } from '@nestjs/common';
 import { UsersService } from 'src/users/users.service';
 import type { AuthSocket } from 'src/auth/jwt-auth.guard';
 
+const rawOrigins = process.env.CORS_ORIGINS ?? '';
+const parts = rawOrigins.split(',');
+
+const trimmedOrigins = parts.map(function (origin) {
+  return origin.trim();
+});
+
+const allowedOrigins = trimmedOrigins.filter(function (origin) {
+  return origin !== '';
+});
+
 @WebSocketGateway({
   cors: {
-    origin: 'http://localhost:1024',
+    origin: allowedOrigins,
     credentials: true,
   },
 })
@@ -27,7 +38,10 @@ export class GameGateway implements OnGatewayConnection, OnGatewayDisconnect {
   @WebSocketServer()
   server!: Server;
 
-  constructor(private readonly gameService: GameService, private readonly usersService: UsersService) {}
+  constructor(
+    private readonly gameService: GameService,
+    private readonly usersService: UsersService,
+  ) {}
 
   handleConnection(client: Socket) {
     console.log(`Client connected : ${client.id}`);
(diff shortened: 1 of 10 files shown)


frontend/src/pages/Home.tsx
```diff
@@ -1,15 +1,15 @@
...
+import { Button } from "@/components/ui/Button";
 
 export default function Home() {
   const { t } = useTranslation();
   const navigate = useNavigate();
 
   const handleFindOpponent = async () => {
     try {
-      const response = await fetch("http://localhost:1024/api/game/create", {
+      const response = await fetch("/api/game/create", {
```
Mikan95: Using fetch here may be at the origin of the 401 issues you mentioned in your PR description.

see #172 where I explain the difference between fetch and using axios! 👍 
For consistency and auth issues I think we should use the axios here like in other areas of the code!

Let me know what you think!

backend/src/modules/game/game.gateway.ts
Mikan95: `AuthSocket` is a TypeScript type that extends the base `Socket` type by declaring that `client.data.user` exists and has a known shape (user ID, email, etc.). 
At runtime that data is attached to the socket regardless of which we use, so both `Socket` and `AuthSocket` actually have `client.data.user` on them when the code runs.

However, at compile time it is different:

// typed as Socket — TypeScript thinks `client.data.user` doesn't exist
client.data.user.sub  // ← TypeScript error, even though it works at runtime

// typed as AuthSocket — TypeScript knows `client.data.user` is there
client.data.user.sub  // ← no problem

So, no bug at the moment as neither handler touches `client.data.user`.

But if later we add a check like `client.data.user.sub` to verify the player is actually in that game, TypeScript will error because it thinks `client.data.user` doesn't exist on`Socket`. 

Should we change this to use `AuthSocket` like `handleJoinGame` does?

Mikan95: Same as for `handleJoinGame` -> change to AuthSocket?

Mikan95 (changes requested):
Good work getting all the routes and stuff working!

I left a couple of comments. Let me know what you think!

frontend/src/pages/Home.tsx
simo1616: Thanks, that makes sense.

I replaced the standalone `fetch` with the shared axios-based service so this call now stays aligned with the rest of the frontend HTTP/auth behavior (`baseURL: /api/` + `withCredentials: true`).

That should also avoid the auth inconsistency I mentioned in the PR description.

backend/src/modules/game/game.gateway.ts
simo1616: Yep, agreed.

I changed this handler to `AuthSocket` as well so it stays aligned with `handleJoinGame`.

It works the same at runtime right now, but this gives us the proper authenticated socket typing and avoids future TypeScript issues if we start using `client.data.user` here.

simo1616: Agreed, updated this one too.

`request_replay` now uses `AuthSocket` for the same reason: consistency with `handleJoinGame` and no future typing issues around `client.data.user`.

Mikan95 (approved):
Looks great! Thanks for the changes to axios and AuthSocket!

PR #176, opened 2026-04-16
# [INFRA] - Separate dev and prod into independently managed stacks

### LINKED ISSUE #173: [INFRA] - Separate dev and prod into independently managed stacks
## Description

The dev and prod builds are currently managed as a single stack, meaning switching between them requires tearing down one and fully rebuilding the other. This makes side-by-side comparison cumbersome and slow.

This issue proposes separating them into two independently managed Docker Compose stacks using project namespacing, so both can run simultaneously on the same machine without conflict.

## Goal

- Eliminate the need to destroy and rebuild one stack just to test the other
- Allow dev and prod to run at the same time on different ports for direct side-by-side comparison
- Make stopping/starting each stack fully independent

## Scope

- [ ] Frontend
- [ ] Backend
- [ ] Database
- [x] Docker / Infrastructure
- [ ] Documentation
- [ ] Other

## Proposed Tasks

- [x] Namespace dev and prod as separate Compose projects
- [x] Assign dedicated host ports to each stack (no overlap)
- [x] Give containers predictable names per stack (`dev-*` / `prod-*`)
- [ ] Test that both stacks can run simultaneously without conflict

## Notes

**After this change:**
Certain targets of make may need two variants to target the specific stack. For example:

```bash
make up    # start dev stack  → http://localhost:1024
make prod  # start prod stack → https://localhost:8443

make down       # stop dev only  — prod keeps running
make down-prod  # stop prod only — dev keeps running
```

**Containers will be namespaced per stack:**

| Stack | Containers |
|---|---|
| dev | `dev-frontend`, `dev-backend`, `dev-proxy`, `dev-db` |
| prod | `prod-frontend`, `prod-backend`, `prod-proxy`, `prod-db` |

Volumes and networks are also isolated — no shared state between stacks.

> **Note:** This may mean that you will need to manually remove previous containers that were already built before merging this change in to your active workspace.

Mikan95:
Closes #173 

## Checklist

- [x] Target branch is `dev`

## What was done

**Infrastructure:**

- Added `-p dev` / `-p prod` project flag to all `docker compose` calls in `Makefile` and `make/clean.mk`
- Removed `container_name` from all services in `compose.yaml`
- Added `container_name: dev-*` for all services in `compose.dev.yaml`
- Added `container_name: prod-*` for all services in `compose.prod.yaml`
- Moved `ports: "1024:80"` from `compose.yaml` to `compose.dev.yaml` so prod does not inherit the dev port
- `make setup` now automatically sets `DOMAIN=127.0.0.1` in `.env`

**Makefile:**

- `make help` reworked to reflect the new target layout
- Added production counterparts for all dev targets: 
  - `p-down`
  - `p-downv`
  - `p-ps`
  - `p-logs`
  - `p-logs-proxy`
  - `p-logs-front`
  - `p-logs-back`
  - `p-logs-db`

## Why

Both stacks can now run simultaneously on different ports — dev on `:1024`, prod on `:8080`/`:8443`. This removes the need to tear down and rebuild one stack to test the other, making it easy to compare behaviour or verify that a change works correctly in both environments.

## Testing

- [x] Docker build successful

## Production Tests

> Make sure `DOMAIN` is set to `127.0.0.1` in `.env` before starting prod.

### Prod Test 1 — Both stacks run simultaneously

**1. Start the prod stack:**

```bash
make prod
```

**2. Start the dev stack:**

```bash
make up
```

**3. Verify both stacks are up:**

```bash
make ps
```

> **Expected:**
>
> ```
> dev-frontend   Up
> dev-backend    Up
> dev-proxy      Up
> dev-db         Up
> ```

```bash
make p-ps
```

> **Expected:**
>
> ```
> prod-frontend  Up
> prod-backend   Up
> prod-proxy     Up
> prod-db        Up
> ```

---

### Prod Test 2 — Stacks are isolated: stopping dev does not affect prod

**1. Stop the dev stack:**

```bash
make down
```

**2. Verify prod is still running:**

```bash
make p-ps
```

> **Expected:** all 4 prod containers still up.

---

### Prod Test 3 — Prod stack is reachable

**1. Check the prod endpoint responds:**

```bash
curl -Ik https://127.0.0.1:8443
```

> **Expected:** `HTTP/2 200`

**2. Stop the prod stack:**

```bash
make p-down
```

---

### Dev Test 1 — Dev stack is reachable

**1. Start the dev stack:**

```bash
make up
```

**2. Check the dev endpoint responds:**

```bash
curl -I http://localhost:1024
```

> **Expected:** `HTTP/1.1 200 OK`

---

### Dev Test 2 — Stacks are isolated: stopping prod does not affect dev

**1. Start the prod stack:**

```bash
make prod
```

**2. Stop the prod stack:**

```bash
make p-down
```

**3. Verify dev is still running:**

```bash
make ps
```

> **Expected:** all 4 dev containers still up.

## Cleanup

With this new setup you'll find that the old containers will not be cleaned up by make nuke anymore. So you may have lingering images and volumes. If that's the case do the following.

### Cleanup Lingering Images

**1. Identify the name of lingering images:**

```bash
make ls
```

**2. Run the following command to remove them manually:**

```bash
docker image rmi image_name_1 image_name_2 image_name_3
```

> **Note:** replace image names with the relevant names that are listed from the ls command.
> You should be able to auto-complete the names using `tab`

In my case the command looked like this, but it will depend what the name of your root directory is:

```bash
docker image rmi transcendence-backend:latest transcendence-frontend:latest transcendence-proxy:latest
```

### Cleanup Lingering Volumes

**1. Identify the name of lingering volumes:**

```bash
docker volume ls
```

**2. Run the following command to remove them manually:**

```bash
docker volume rm volume_name_1 volume_name_2 volume_name_3 etc...
```

> **Note:** the volumes you want to keep are below. All others can safely be removed.

```text
DRIVER    VOLUME NAME
local     dev_backend_node_modules
local     dev_frontend_node_modules
local     dev_postgres_data
local     prod_postgres_data
```

**3. For good measure run a make clean:**

```bash
make clean
```


Makefile
```diff
@@ -14,31 +14,29 @@ SERVICE_DATABASE	= db
...
-no-cache: setup ##  Rebuild all containers in no-cache mode [DEV]
-	@docker compose -f $(COMPOSE_FILE) -f $(COMPOSE_DEV) build --no-cache
+no-cache: setup ## Rebuild all containers in no-cache mode [DEV]
+	@docker compose -p dev -f $(COMPOSE_FILE) -f $(COMPOSE_DEV) build --no-cache
 
-re: down build up ##  Stop, rebuild, and restart the full stack [DEV]
+re: down build up ## Stop, rebuild, and restart the full stack [DEV]
 
-reset: down-v no-cache up ##  Stop (remove volumes), full rebuild (no cache), restart containers [DEV]
+reset: down-v no-cache up ## Stop (remove volumes), full rebuild (no cache), restart containers [DEV]
```
simo1616: I think this should be `downv` instead of `down-v`.

I tested it locally and `make reset` currently fails because `down-v` no longer exists.

simo1616:
I tested the main behavior locally and confirmed that the dev/prod stack separation works as intended:

* both stacks can run simultaneously
* dev and prod are reachable on their respective ports
* stopping one stack does not affect the other

However, I also reproduced the issue mentioned on `reset`:

```bash
make reset
```

currently fails because it still depends on `down-v`, while the target was renamed to `downv`.

Once that target is fixed, this looks good to me.

Makefile
Mikan95: Well spotted thanks you! I'll make the change now!
I've also added a **Cleanup** section to the PR desscription to guide you how to cleanup old containers and volumes as they won't be deleted by `make nuke` anymore!

Mikan95:
> However, I also reproduced the issue mentioned on `reset`:
> 
> ```shell
> make reset
> ```
> 
> currently fails because it still depends on `down-v`, while the target was renamed to `downv`.


Good catch! Thanks I've updated this now so should be good! Make sure to checkout the `Cleanup` section I added to the PR description for instructions on how to remove old containers and volumes :+1:

simo1616 (approved):
Thanks for the quick fix on make reset. Everything is working as intended now, and thanks for the cleanup guide too. Approved!

PR #177, opened 2026-04-17
[FRONTEND] User Profile Page, Match History & XP Progression System

### LINKED ISSUE #127: [BACKEND] Match History & Win-Ratio API Endpoints
### Description 
- Create the endpoints necessary for the frontend to display a user's competitive past and performance metrics.

### Tasks

- [x] Implement GET /users/:id/matches to retrieve the list of past games for a specific user.

- [x] Implement GET /users/:id/stats to return calculated metrics (Win/Loss ratio, total games played).

- [x] Optimize the Prisma query to include the usernames of opponents in the match history list.

nasdhn:
## 📋 Summary

This PR introduces the **Match History tracking system**.  
It implements a player progression model (XP/Levels) and ensures that game results are permanently archived in the database using an atomic scoring model.

---

## 🔗 Linked Issues

Closes #127 

---

### 🎮 Progression
XP-based leveling system with balanced rewards per match outcome.

| Result | XP Gained | Level Formula |
|---|---|---|
| **Win** | +20 XP | `level = xp / 100` |
| **Draw** | +10 XP | `level = xp / 100` |
| **Loss** | +5 XP | `level = xp / 100` |

### 🗄️ Match History
Atomic match recording — each game is saved as a clean `1-0` or `0-1` result, not a cumulative session score.

### 👤 Profile UI

---

### Frontend (React)

- **`Profile.tsx`** — Added a Match History section
- **XP Progress Bar** — Visual bar with level calculation (`level = xp / 100`)
- **Match History List** — Displays date, result (WIN / LOSS / DRAW), and opponent details

### Backend (NestJS & Prisma)

- **`MatchesService`**
  - `getFinishedGamesHistory` — returns a formatted list of past games for the profile page

- **`GameService`**
  - `saveGameToDB` now normalizes scores to `1-0` or `0-1` before writing to the DB — prevents cumulative scores from corrupting stats

---

### Match History Logic — `getFinishedGamesHistory`

This section introduces the logic that transforms raw database records into a readable, user-contextual match history.

---

#### 🧠 How it works

**Bidirectional fetch** — The query retrieves all matches where the user appears, whether they were recorded as `player1` or `player2`:

```typescript
where: {
  OR: [
    { player1Id: userId },
    { player2Id: userId },
  ],
},
```

**Dynamic opponent resolution** — The backend automatically determines which profile (`player1` or `player2`) is the opponent, and returns a single unified `opponent` object to the frontend. No ambiguity on the client side.

**Contextual result calculation** — `result` (WIN / LOSS / DRAW) is computed dynamically by comparing the current user's ID against the stored `winnerId`:

| Condition | Result |
|---|---|
| `winnerId === userId` | `WIN` |
| `winnerId === null` | `DRAW` |
| anything else | `LOSS` |

**Score inversion** — `myScore` and `oppScore` are swapped based on the user's role in the match (P1 or P2), ensuring the score always reads from the user's perspective.

---

#### 📦 Response shape

Each entry in the history returns the following structure:

```typescript
{
  id: number;
  date: Date;
  result: "WIN" | "LOSS" | "DRAW";
  myScore: number;
  oppScore: number;
  opponent: {
    username: string;
    avatar: string;
  };
}
```

---

> ⚙️ **Performance note:** The Prisma query uses `include` to fetch full player profiles and winner information in a **single database call** — no N+1 queries, keeping the profile page load fast regardless of match history size.

---

### Test 1 — Profile Display

Navigate to `/profile`, refresh.

**Expected:**
- Wins / Losses / Draws stats are visible
- XP bar shows correct progress (e.g. `40/100 XP` after 2 wins)
- Level is calculated correctly (`level = xp / 100`)

---

### Test 2 — XP & Stats Update After a Match

Play a full match until the end, then go to `/profile`, refresh.

**Expected:**

| Outcome | XP gained | Counter updated |
|---|---|---|
| Win | +20 XP | `Wins +1` |
| Draw | +10 XP | `Draws +1` |
| Loss | +5 XP | `Losses +1` |

---

### New Endpoint — Match History API

`GET /api/users/:id/history` — returns the full formatted match history for a given user. Protected by `JwtAuthGuard`.

---

#### 🧪 How to Test

**Step 1 — Get your token**

Open the app in your browser, log in, then open `F12` → `Application` → `Cookies` → copy the value of the `access_token` cookie.

**Step 2 — Get your user ID**

```bash
curl -i -b "access_token=YOUR_TOKEN_HERE" http://localhost:1024/api/auth/me
```

Copy the `id` field from the response.

**Step 3 — Call the history endpoint**

```bash
curl -i -b "access_token=YOUR_TOKEN_HERE" http://localhost:1024/api/users/YOUR_ID_HERE/history
```

---

#### 📤 Expected Response

An array of match objects — each entry represents a single atomic match. Scores are always `1-0` or `0-1`, and `result` is computed relative to the requested user.

```json
[
  {
    "id": 12,
    "date": "2026-04-17T08:58:02.161Z",
    "result": "WIN",
    "myScore": 1,
    "oppScore": 0,
    "opponent": {
      "username": "Mechant",
      "avatar": "default.png"
    }
  },
  {
    "id": 10,
    "date": "2026-04-17T08:57:28.628Z",
    "result": "LOSS",
    "myScore": 0,
    "oppScore": 1,
    "opponent": {
      "username": "Mechant",
      "avatar": "default.png"
    }
  }
]
```

🚨 **If you get `401 Unauthorized`** → your token is invalid or expired, log out and log back in to get a fresh one.  
🚨 **If you get an empty array `[]`** → the user has no finished matches yet, play a full game first.

---

### UI — Minimalist Match History

> 🎨 **Design note:** The current profile UI is intentionally bare-bones — data is displayed in a near-raw format. A dedicated pass will be done to make it visually polished. The API already sends everything needed, so the frontend work will be purely cosmetic.

Following team feedback, scores were removed from the primary match history list. The view focuses on **Result** and **Opponent** only, keeping the UI clean.  
The full score data remains available in the API for a future detailed match view.


frontend/src/pages/Profile.tsx
```diff
@@ -68,6 +78,26 @@ export default function Profile() {
...
     }
 
+    useEffect(() => {
+      if (user) {
+        setUserName(user.username);
+        setBio(user.bio);
+
+        async function fetchhistory() {
+          try {
+            const data = await userService.getUserHistory(user.id);
```
simo1616: Nice work here, the flow is working well in dev from what I tested

Small thing to fix so it also passes in prod build: TypeScript can still see `user` as possibly `null` when `user.id` is used here.

On my side it breaks with:

src/pages/Profile.tsx(88,59): error TS18047: 'user' is possibly 'null'.

Snippet:
const data = await userService.getUserHistory(user.id);

So in dev it can still run, but in prod `npm run build` / `tsc -b` blocks here. I think this just needs a small restructure so TS can clearly understand that `user` exists at this point.

simo1616:
Really nice work on this one — I tested the feature locally after patching the current Home import issue on my side, and the main flow is working well:

- XP progression updates correctly
- match history is returned correctly
- profile display makes sense from both users' perspectives

So the feature direction looks good to me.

I’m not approving yet only because there is still one real blocker for the branch: the frontend prod build fails in `Profile.tsx`.

What I got locally:

`npm run build`
-> `src/pages/Profile.tsx(88,59): error TS18047: 'user' is possibly 'null'.`

`npx tsc -b`
-> same error

`npm run lint`
-> also reports:
- `React Hook "useEffect" is called conditionally`
- plus a few existing `any` warnings/errors around the file

`docker compose --progress plain -p prod -f compose.yaml -f compose.prod.yaml build frontend`
-> same build failure during the frontend image build

So from my side, the remaining thing before approval is mainly:
- make `Profile.tsx` pass cleanly in build / TS / prod
- and clean up the hook placement / nullable `user` handling there

Small separate note: the current `Home.tsx` import mismatch is on me / inherited branch noise, not on this PR. I’m not counting that against your feature. For local compilation on the current branch state, it may need:
`import { gameApi } from "@/services/gameApi";`
instead of
`import { gameApi } from "@/services/gameService";`

For the history endpoint visibility / access-control side, I think that one is more of a team decision than a hard review blocker from me right now, so I’ll raise it separately as a technical discussion.

Overall though: very solid progress, and the feature itself is clearly moving in the right direction.

simo1616:
I re-checked the subject and I didn’t find anything explicitly forcing match history to be private or visible to other authenticated users.

So overall, this doesn’t seem like required work from the subject for now. I’d see it more as an optional team decision for later than something we need to add right away.

frontend/src/pages/Profile.tsx
nasdhn: Thanks for the feedback! I've addressed the TS issue in Profile.tsx. I restructured the component to respect the React Hook order and added a non-null assertion for the user ID to ensure the production build passes. Should be good to go now!

simo1616: Re-tested on my side in prod-local, this looks fixed now. Resolving. thanks @nasdhn

simo1616:
Re-tested in prod-local on my side and the previous Profile.tsx build blocker looks fixed now.

Validated:
- prod build/start works
- profile page works in prod
- match history endpoint works
- XP/stats/history display correctly
- both users see coherent results from their own perspective

Nice work, approving.

simo1616 (approved):
Re-tested in prod-local on my side and the previous Profile.tsx build blocker looks fixed now.

Validated:
- prod build/start works
- profile page works in prod
- match history endpoint works
- XP/stats/history display correctly
- both users see coherent results from their own perspective

Nice work, approving.

PR #183, opened 2026-04-18
[GAME][REMOTE] Preserve player identity across disconnect/reconnect

### LINKED ISSUE #179: [GAME][REMOTE] Preserve player identity across disconnect/reconnect
## Goal
Allow a remote player to recover their seat and role after a socket disconnect or page refresh.

## Context
The current game flow is still heavily tied to the live socket connection.
For remote play, the server should identify the player independently from the current socket id, so a reconnect does not break seat ownership.

## Tasks
- [x] Introduce a stable player identity key for game ownership
- [x] Stop treating `socket.id` as the long-term player identity
- [x] Update player seat assignment to use the stable player identity
- [x] Allow a reconnecting player to recover `X` or `O`
- [x] Resync the current game state after reconnect
- [x] Keep behavior stable for the normal join flow

## Definition of Done
- [x] A player who refreshes or reconnects can recover the same role
- [x] The current game state is resynced correctly
- [x] Seat ownership is no longer lost just because the socket changed
- [x] Normal two-player join flow still works

simo1616:
## # Summary
This PR stops game seat ownership from depending on `socket.id` alone. 
Seats are now owned by **authenticated users**, allowing a player to refresh or reconnect and recover their `X` or `O` role instead of losing ownership due to a new socket connection.

---

---

## # Why
On the current `dev` branch, seat ownership is tied to raw Socket.IO connection IDs. A page refresh makes the user look like a "new" player, breaking the game state.
This branch separates **stable ownership** (User ID) from **live presence** (Socket ID).

---

### Backend
- **Stable Seats:** Replaces raw socket strings with a `PlayerSeat { ownerUserId, socketId }` structure.
- **Seat Recovery:** `assignPlayerRole()` now checks if the `userId` already owns a seat before creating a new one.
- **Persistence:** Disconnect no longer deletes the player; it sets `socketId = null` and triggers a grace period.
- **Move Timer Reset:** If the reconnecting player is the `currentPlayer`, `lastMove` is reset to prevent immediate timeout after the loading screen.
- **Grace Period:** Added a reconnect timer in the Gateway. If it expires while the opponent is still online, the game ends by `forfeit`.

### Frontend
- **Types:** Updated `PlayerSeat` to match the backend.
- **Logic:** Minimal changes; state resync still relies on the existing `joined_as` + `game_updated` flow.

---

## # Key Files to Review
1. `backend/src/modules/game/game.service.ts` (Core logic & forfeit)
2. `backend/src/modules/game/game.gateway.ts` (Timers & socket management)
3. `backend/src/modules/game/game.logic.ts` (Seat assignment)
4. `backend/src/modules/game/game.types.ts` & `frontend/src/type/game.types.ts`

---

### Disconnect -> Reconnect / Forfeit Flow
```mermaid
sequenceDiagram
    participant X as Player X
    participant O as Opponent
    participant G as Gateway
    participant S as Service
    participant T as Reconnect timer

    X-xG: disconnect
    G->>S: processPlayerDisconnection(socketId)
    S-->>G: socketId = null
    G->>T: start grace timer

    alt reconnect before timer expires
        X->>G: join_game(gameId)
        G->>S: joinGame(gameId, newSocketId, sameUserId)
        S-->>G: restore X
        G->>T: clear timer
        G-->>X: joined_as X
        G-->>All: game_updated
    else no reconnect
        T-->>G: timer expires
        G->>S: finalizeReconnectTimeout(gameId, X)
        S-->>G: winner = O, endReason = forfeit
        G-->>All: game_updated
    end
```

---

## # Code Highlights

**Stable seat structure:**
```typescript
export interface PlayerSeat {
  ownerUserId: number | null;
  socketId: string | null;
}
```

**Reset move timeout on active-player reconnect:**
```typescript
if ((isReconnectX || isReconnectO) && game.status === 'playing' && role === game.currentPlayer) {
  game.lastMove = Date.now();
}
```

---

## # How to Test (Manual)
1. **Refresh Recovery:** Start a game between two logged-in users. Refresh the page.
   - *Expected:* You should automatically rejoin as the same role (X or O).
2. **Move Timeout Reset:** Make it Player A's turn. Refresh Player A.
   - *Expected:* Player A returns, and the turn timer effectively restarts from the moment of reconnection.
3. **Forfeit:** Disconnect Player A and wait past the grace period while Player B stays online.
   - *Expected:* Game ends, Player B wins by `forfeit`.
4. **Resync:** Play 5 moves, then refresh.
   - *Expected:* The board state is perfectly restored.

---

## # Known Limitations & Risks
- **Auth Only:** Reconnect only works for authenticated users (required for stable `userId`).
- **Both Offline:** If both players disconnect, the game currently does not auto-resolve.
- **UI:** No specific "Player Disconnected" overlay has been added to the frontend yet.

---

## # Reviewer Checklist
- [ ] Seats are stable by `ownerUserId`.
- [ ] Disconnect only clears `socketId`, does not wipe the seat.
- [ ] Reconnect clears the pending forfeit timer.
- [ ] Move timeout resets upon active player return.
- [ ] Forfeit finalizes correctly if the opponent is online.

Close #179

### DIFF
--- a/backend/src/modules/game/game.gateway.ts
+++ b/backend/src/modules/game/game.gateway.ts
@@ -40,28 +40,47 @@ export class GameGateway implements OnGatewayConnection, OnGatewayDisconnect {
 
   private timersForfeit = new Map<string, NodeJS.Timeout>();
 
+  constructor(
+    private readonly gameService: GameService,
+    private readonly usersService: UsersService,
+  ) {}
+
   private getTimerKey(gameId: string, role: 'X' | 'O') {
     return `${gameId}:${role}`;
   }
 
   private clearTimerForfeit(gameId: string, role: 'X' | 'O') {
-    const timeKey = this.getTimerKey(gameId, role);
-    const timer = this.timersForfeit.get(timeKey);
+    const timerKey = this.getTimerKey(gameId, role);
+    const timer = this.timersForfeit.get(timerKey);
 
     if (timer) {
       clearTimeout(timer);
-      this.timersForfeit.delete(timeKey);
+      this.timersForfeit.delete(timerKey);
     }
   }
-  private startReconnectTime(gameId: string, role: 'X' | 'O') {
-    const timeKey = this.getTimerKey(gameId, role);
-    this.clearTimerForfeit(gameId, role); // clear the last one
-  }
 
-  constructor(
-    private readonly gameService: GameService,
-    private readonly usersService: UsersService,
-  ) {}
+  private startReconnectTimer(gameId: string, role: 'X' | 'O') {
+    const timerKey = this.getTimerKey(gameId, role);
+    this.clearTimerForfeit(gameId, role);
+
+    const timer = setTimeout(() => {
+      this.gameService
+        .finalizeReconnectTimeout(gameId, role)
+        .then((result) => {
+          if (result) {
+            this.server.to(gameId).emit('game_updated', result.game);
+          }
+        })
+        .catch((error) => {
+          console.error('Reconnect timeout error:', error);
+        })
+        .finally(() => {
+          this.timersForfeit.delete(timerKey);
+        });
+    }, 45000);
+
+    this.timersForfeit.set(timerKey, timer);
+  }
 
   handleConnection(client: Socket) {
     console.log(`Client connected : ${client.id}`);
@@ -72,7 +91,7 @@ export class GameGateway implements OnGatewayConnection, OnGatewayDisconnect {
     const result = this.gameService.processPlayerDisconnection(client.id);
     if (!result) return;
     if (result.game.status === 'playing')
-      this.startReconnectTime(result.gameId, result.role);
+      this.startReconnectTimer(result.gameId, result.role);
     this.server.to(result.gameId).emit('game_updated', result.game);
   }
 
--- a/backend/src/modules/game/game.gateway.ts
+++ b/backend/src/modules/game/game.gateway.ts
@@ -55,6 +55,7 @@ export class GameGateway implements OnGatewayConnection, OnGatewayDisconnect {
 
     if (timer) {
       clearTimeout(timer);
+      console.log(`[RECONNECT] cleared timer for ${role} in game ${gameId}`);
       this.timersForfeit.delete(timerKey);
     }
   }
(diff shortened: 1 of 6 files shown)


simo1616:
Quick update after the `dev` merge conflict resolution.

I resolved the conflict in `backend/src/modules/game/game.service.ts` by keeping the reconnect logic from this PR, while also preserving the newer match-save behavior from `dev`.

So the final merged behavior is now:

* seats stay attached to a real authenticated user through `PlayerSeat { ownerUserId, socketId }`
* a disconnect no longer ends the game immediately
* the reconnect flow still uses the grace-period / delayed-forfeit logic from this PR
* match persistence keeps the atomic save behavior from `dev` (`1-0`, `0-1`, `0-0`) instead of saving cumulative session scores

One important adjustment after reviewing the behavior more carefully: the previous reconnect snippet that reset the move timer on active-player reconnect was removed:

```ts
if ((isReconnectX || isReconnectO) && game.status === 'playing' && role === game.currentPlayer) {
  game.lastMove = Date.now();
}
```

In the end, this was not the right behavior for the game flow.

The move timer and the reconnect timer should stay as two separate things:

* the **30s game timer** is the real turn limit and should keep running even if the player disconnects and comes back
* the **20s reconnect grace period** is only the maximum time allowed to stay disconnected before being considered forfeited

So if a player disconnects after already spending part of the 30s turn, that elapsed time should still count when they return. Reconnecting should restore the seat and the game state, but it should not grant extra thinking time by restarting the move timer.

Example:

* player disconnects after 5s of their turn
* comes back before the reconnect grace period expires
* they should resume with ~25s left on the turn, not a fresh 30s

That felt more consistent and more fair for the opponent, so the branch now follows that behavior.

Sorry for the delay on this one, and thanks again for the review.

nasdhn (approved):
Hello everyone,

I've completed the manual validation of the reconnection logic on the `feature/game-reconnect` branch. The implementation is now stable and ready for merge. Here is my report:

### Validation Summary
I performed the following manual tests:
* **Refresh Recovery:** Successfully recovered the seat ownership via `userId` across multiple refreshes.
* **Move Timeout Reset:** Verified that the timer correctly resets upon reconnection for the active player.
* **Refresh Abuse Fix:** I've confirmed that the infinite refresh exploit has been resolved; players can no longer abuse page refreshes to reset their turn timer indefinitely.
* **Forfeit Flow:** Disconnected the player and waited for the grace period; the opponent correctly won by forfeit.
* **Board Resync:** The game state is perfectly restored after a page refresh.

### 💡 Improvements & Edge Cases for Follow-up
While the core logic is solid and safe, I've identified a few areas we should address in future iterations:

1. **Memory Management (RAM):** Finished game sessions currently remain in server memory indefinitely. We should implement a cleanup strategy

2. **Double Inactivity Logic:** I noticed that if two players launch a game but neither plays, the system currently attributes a win to one of them via timeout. We should investigate this to handle double-inactivity cases properly (e.g., declaring a draw or a no-contest instead of a random win).

Great work!

PR #222, opened 2026-04-26
[INFRA] - Seed database with dummy users and simulated game history

### LINKED ISSUE #206: [INFRA] - Add dummy users to Database
## Description
During this week's meeting we discussed the possibility of adding dummy users to the database. To avoid devs needing to create users/play games manually after each rebuild.

## Goal

- Add multiple dummy users with varying usernames, number of matches played, wins and losses.

## Scope

- [ ] Frontend
- [ ] Backend
- [x] Docker / Infrastructure
- [x] Database
- [ ] Documentation
- [ ] Other

Mikan95:
## Previously Blocked by: Achievement Seeding Discussion

While working on `seed.ts` it became apparent that achievement definitions (e.g. `FIRST_WIN`) had to be inserted into the `Achievement` table on every fresh stack before any achievement could be unlocked, because `UserAchievement` held a foreign key to it. Forgetting to seed meant unlocks silently failed with no error.

**Decision: achievement definitions will be moved to a TypeScript constant in the backend.** The database will store only user unlock records (`key` string + timestamp), removing the seeding dependency entirely. 

**Brief reasons:**

- The seeded approach required two places to stay in sync: `seed.ts` and the service. Adding an achievement meant updating both; they could silently drift.
- Keys were plain strings — a typo (`'FIRST_GAM'`) would cause unlocks to silently do nothing at runtime instead of failing at compile time.
- Each unlock required 3 DB round-trips (resolve key → check duplicate → write) instead of 1 upsert.

> **Tradeoff:** Achievement display names and descriptions are no longer queryable from raw SQL. A `SELECT * FROM "UserAchievement"` gives keys only — display data now lives in the backend constant and must be cross-referenced in code. For a project of this scale this is an acceptable tradeoff.

---

EDIT:
Previously also blocked by issues related to the revert commit `53acb13` which have since been resolved.

**The following actions were undertaken to ensure a clean and linear git history:**

- `git rebase dev` to avoid reinjecting problem commit
- `git push --force-with-lease origin infra/build-targets-improvements` to avoid duplicate commits

---

closes #206

## Checklist

- [x] Target branch is `dev`

### **Backend:**

- Added `prisma/seed.ts` — seeds 10 dummy users (`dummy0`–`dummy9`) with hashed passwords, auto-generated dicebear avatars, and 20 simulated games between them
- Computes and writes `wins`, `losses`, `draws`, and `xp` onto each user after games are inserted
- Extracted `DUMMY_USERNAMES` constant used for all dummy-scoped queries — avoids `startsWith` matching unintended usernames (e.g. a real user named `dummyjohn`)
- On re-run, wipes only games where **both** players are dummy accounts — games between a dummy and a real user are preserved
- Added `prisma/seed.ts` to the exclude list in `tsconfig.build.json` to prevent TypeScript from treating it as part of the production build

### **infra:**

- Added `seed` property to `prisma.config.ts` pointing to `ts-node ./prisma/seed.ts`
- Added `make seed` target to `make/setup.mk` — runs `npx prisma db seed` inside the running backend container
- `docker-entrypoint.sh` runs the seed automatically on every container start — dummy data is always present after `make up`

## Why

Developing and testing features like leaderboards, match history, and user profiles requires realistic data in the database. Manually creating users and games through the UI each time the stack is reset is slow and error-prone. `make seed` solves this by populating the DB with a consistent, reproducible dataset in one command.

> **Demo warning:** Because the seed runs on every container start, any game played between two dummy accounts will be wiped on the next `make down` / `make up`. If you need game data to persist across stack restarts during a demo, create a real user account and play against a dummy — those games are out of scope for the wipe.

## Testing

> This is a dev-only feature. No production tests required.

### Dev Test 1 — Seed runs cleanly

**1. Start the dev stack:**

```bash
make up
```

**2. Run the seed:**

```bash
make seed
```

> Expected output: `Seeded dummy0` … `Seeded dummy9`, `Added Games to Database`, `Updated Users' XP and Stats`

**3. Verify users and stats were written:**

```bash
docker exec dev-db psql -U postgres_superuser -d transcendence -c 'SELECT username, wins, losses, draws, xp FROM "User" ORDER BY xp DESC;'
```

> Expect 10 rows with non-zero wins/losses/xp.

**4. Verify games were written:**

```bash
docker exec dev-db psql -U postgres_superuser -d transcendence -c 'SELECT id, "player1Id", "player2Id", "winnerId", "scoresP1", "scoresP2", "endReason" FROM "Game";'
```

> Expect 20 rows.

---

### Dev Test 2 — Seed is idempotent

**1. Run `make seed` a second time without resetting the stack.**

**2. Re-run both queries from Dev Test 1.**

> Expect the same 10 users and exactly 20 games (not 40) — dummy-vs-dummy games are wiped and re-inserted on each run.

---

### Dev Test 3 — Real user games are preserved on re-seed

**1. Register a real user account through the UI.**

**2. Play a game between the real user and any dummy account.**

**3. Run `make seed` again.**

**4. Re-run the games query from Dev Test 1.**

> Expect 21 rows — the 20 seeded games plus the real user's game, which was not touched by the wipe.


simo1616:
Yeah I think @Mikan95 has a point here.

The seed thing can be annoying, because after a reset it’s one more thing that has to be correct. And if it’s not, achievements just don’t unlock and it’s not really obvious why.

So yeah, for our case, just storing the key in `UserAchievement` sounds simpler.

But at the same time, Nasim already made the current version and if it works now, I don’t know if it’s worth changing everything just for that. If it’s quick, why not, but if it adds too much work then maybe not.

I think we can keep it like this for now and only simplify it later if the seed becomes a problem again, or if we have enough time.

@nasdhn what do you think?

nasdhn:
I understand the concern about the 'forgotten' seed, that's why I've already automated it in the entrypoint. In practice, the seed starts automatically every time the stack resets, so the risk of error or forgetting is already zero. Refactoring now would waste time for the same result, even though it works.

However, if you really want to simplify things by using constants, I can refactor everything... Let me know what you prefer

Mikan95:
> I understand the concern about the 'forgotten' seed, that's why I've already automated it in the entrypoint. In practice, the seed starts automatically every time the stack resets, so the risk of error or forgetting is already zero. Refactoring now would waste time for the same result, even though it works.
> 
> However, if you really want to simplify things by using constants, I can refactor everything... Let me know what you prefer

Thanks for your reply! @nasdhn 

There are 2 main questions I'd like us to discuss and explore as a team:
**1. Should seed be automatic and thus run for both dev and prod?**
  - Pros: 
    - Negligible changes needed (Removing `make seed` command).
    - Having dummy users in prod will be useful for the demonstration during eval
  - Cons:
    - No flexibility, stuck with dummy users no matter what.

**2. Should the achievements be seeded into the DB at all?**

Obviously, I understand at this stage it may be too late and refactoring the whole achievement system could be time consuming and frustrating. Let's discuss this during tomorrow's meeting and get the whole team's opinion! :+1:

Nico-Ry (approved):
## ✅ Review & Testing

Great work on this feature — this is a really solid addition to the dev workflow and makes testing much easier.

I went through the PR description and tested all the scenarios you listed:

- `make up` and `make seed` both run cleanly
- 10 dummy users are created with consistent stats
- 20 games are inserted correctly
- Re-running the seed keeps the dataset stable (no duplication)
- Real user vs dummy games are preserved as expected

Everything behaved exactly as described and I consistently got the expected results 👍

---

## ❓ Small Observation

One thing I’m not 100% sure I understood correctly:

After running `make nuke` and then `make up`, the dummy users and data already seem to be present without explicitly running `make seed`.

From the PR description, I assumed seeding would be triggered either manually (`make seed`) or via the container startup script — so this might be expected behavior if the seed is intentionally run on container start.

Just wanted to flag it in case it’s unintended, but in any case this is not blocking for this PR.

---

## 🚀 Conclusion

Feature works as intended and integrates well with the current setup.

Really nice job on making the dataset reproducible and safe (especially the selective wipe logic 👌)

Approving and closing ✅

PR #241, opened 2026-05-01
[BACKEND][USERS] Secure user API and profile updates

Nico-Ry:
## Summary

This PR secures user-related endpoints and improves authentication rules.

### User API security

* Removed `email` from public user responses
* Introduced a mapping layer (`toPublicUser`)
* Ensured `/users` only returns safe public profile data

### Authorization

* Restricted `PATCH /users/:id` so users can only update their own profile
* Added check using JWT payload (`req.user.sub`)

### Password security

* Enforced stronger password rules:

  * minimum 8 characters
  * at least one lowercase letter
  * at least one uppercase letter
  * at least one number
* Applied validation to both register and login DTOs

## Behavior

* `/users` no longer leaks sensitive data
* Users cannot modify other users’ profiles
* Weak passwords are rejected at registration and login

## Limitations

* Updating `email` is currently not allowed
* This will be handled in a future PR with a proper secure flow

## Tests

* Verified `/users` no longer returns email
* Verified `/users/:id` returns only public fields
* Verified `PATCH /users/:id`:

  * works for own user
  * returns 403 for other users
* Tested password validation:

  * weak password → rejected
  * strong password → accepted
* Tested login with valid credentials

### Swagger testing

* Used Swagger (`/api/api-docs`) to test endpoints
* Modified request bodies directly in Swagger before execution
* Observed responses and status codes to validate behavior

## Notes

Email update and related account changes will be handled in a separate PR.

Closes #239

### DIFF
--- a/backend/src/auth/dto/login.dto.ts
+++ b/backend/src/auth/dto/login.dto.ts
@@ -1,5 +1,5 @@
 import { ApiProperty } from '@nestjs/swagger';
-import { IsEmail, IsString, MinLength } from 'class-validator';
+import { IsEmail, IsString, Matches, MinLength } from 'class-validator';
 
 export class LoginDto {
 		@ApiProperty({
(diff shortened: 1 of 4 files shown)


simo1616 (approved):
Approved, tested locally.

I checked the user endpoint security changes through Swagger.

Verified that `/users` no longer exposes email and only returns public profile data.
Also checked `/users/:id` and the response looks safe.

Tested `PATCH /users/:id`:
- updating my own user works
- trying to update another user returns `403`

I also tested the password validation:
- weak passwords are rejected
- strong passwords are accepted
- login still works with valid credentials

The limitation about email update makes sense to me, it is better to handle that later with a proper secure flow.

Looks good to me.

PR #244, opened 2026-05-02
[BACKEND][USERS] Add typed user mapping and safer user queries

### LINKED ISSUE #243: [BACKEND][USERS] Add typed public/private user response mapping
## Goal
Make user API responses safer and easier to maintain by replacing `any` with typed response mappers.

## Problem
`toPublicUser(user: any)` works, but `any` disables TypeScript checking. This is risky because user data contains private fields like `email`, `passwordHash`, `twoFactorSecret`, and `twoFactorTempSecret`.

## Tasks
- [ ] Replace `any` in `toPublicUser`
- [ ] Keep public user response unchanged
- [ ] Prepare clean structure for a future `toPrivateUser`
- [ ] Test `GET /users`, `GET /users/:id`, and `PATCH /users/:id`

## Constraints
- Do not change frontend response shape
- Do not expose private DB fields
- Do not add pagination in this issue
- Do not refactor history yet

## Testing
- Swagger
- Browser
- Optional curl

Nico-Ry:
## Summary

This PR improves the backend user API architecture by adding typed user response mapping, limiting Prisma queries to the fields actually needed, and adding simple pagination to the users endpoint.

## What was done

- Added typed public user response mapping in `UsersService`
- Added typed private user response mapping in `AuthService`
- Updated public user queries to use Prisma `select`
- Updated `/auth/me` and register responses to use a private user mapper
- Narrowed match history user relation queries to avoid loading full related users
- Added `limit` and `offset` pagination to `GET /users`

## Match history note

I changed the match history query to avoid loading full related User objects.

The API response was already safe because it only returned opponent username and avatar, but Prisma was still fetching complete player objects internally through `player1: true`, `player2: true`, and `winner: true`.

The fix keeps the response exactly the same while limiting related users to only the fields needed for history display.

I removed `winner` from the include because the service only uses the scalar `winnerId`, not the related `winner` user object.

## Tests (you can test using Swagger directly or adding /api/... to your URL)

- Tested `GET /api/users`
- Tested `GET /api/users?limit=2`
- Tested `GET /api/users?limit=2&offset=2`
- Tested `GET /api/users/:id`
- Tested `GET /api/auth/me`
- Tested `POST /api/auth/register`
- Tested `GET /api/users/:id/history`
- Confirmed public user responses do not expose email, password hash, or 2FA secrets
- Confirmed private auth response includes email and `isTwoFactorEnabled`, but not password hash or 2FA secrets
- Confirmed match history response shape stays unchanged

## Future work

- Add response DTOs or shared mapper utilities if more services need user response mapping
- Add stronger query parameter validation with DTOs
- Design a separate safe email update flow
- Add rate limiting later

Closes #243

### DIFF
--- a/backend/src/auth/auth.service.ts
+++ b/backend/src/auth/auth.service.ts
@@ -5,6 +5,7 @@ import {
   UnauthorizedException,
 } from '@nestjs/common';
 import { JwtService } from '@nestjs/jwt';
+import { Prisma } from '@prisma/client';
 import { PrismaService } from '../prisma/prisma.service';
 import { RegisterDto } from './dto/register.dto';
 import { LoginDto } from './dto/login.dto';
@@ -25,6 +26,23 @@ export type LoginResult =
 	| FullAuthLoginResult
 	| TwoFactorRequiredLoginResult;
 
+const privateUserSelect = {
+	id: true,
+	email: true,
+	username: true,
+	bio: true,
+	avatar: true,
+	wins: true,
+	losses: true,
+	draws: true,
+	xp: true,
+	isTwoFactorEnabled: true,
+} satisfies Prisma.UserSelect;
+
+type PrivateUser = Prisma.UserGetPayload<{
+	select: typeof privateUserSelect;
+}>;
+
 @Injectable()
 export class AuthService {
 	constructor(
@@ -33,7 +51,7 @@ export class AuthService {
 		private readonly twoFactorService: TwoFactorService,
 	) {}
 
-	private toPrivateUser(user: any) {
+	private toPrivateUser(user: PrivateUser) {
 	return {
 		id: user.id,
 		email: user.email,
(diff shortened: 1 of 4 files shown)


simo1616 (approved):
Approved, good backend cleanup.

I like that the public and private user responses are now separated. It makes the API safer and easier to understand, especially for fields like email, passwordHash and 2FA secrets.

The Prisma select changes also make sense. Even if the response was already safe, it is better to not load full related users when we only need username/avatar for history.

Pagination on GET /users is also a good small improvement.

I tested the main endpoints with Swagger and the response shapes look ok. Match history still looks unchanged, and sensitive fields are not exposed.

Small note: in Swagger, I had to put values for limit/offset, otherwise it does not accept the request. But when I call the API directly without Swagger, GET /api/users works fine without them. So for me it does not block this PR, because the API behavior is okay. It looks more like a Swagger documentation issue.

Another small thought for later: maybe we can think about making /auth/me lighter. Right now it includes wins, losses, draws and xp. It is not a big problem because they are direct user fields, but maybe auth/me should stay focused on session/account info, and game stats could come from /users/:id or a profile/stats endpoint.

For me this PR is good to merge.

PR #249, opened 2026-05-03
[BACK-FRONT] Track Total Games Played — Leaderboard

nasdhn:
## Summary

Adds a `totalGames` counter to each user to enrich profile statistics and enable a new sorting mode in the leaderboard. Instead of computing game counts on the fly at query time, the counter is persisted directly in the database for optimal performance.

---

## Files modified

| File | Description |
|------|-------------|
| `prisma/schema.prisma` | Added `totalGames` field (`Int`, default `0`) to the `User` model |
| `prisma/seed.ts` | Updated seed script to retroactively compute and initialize `totalGames` for test accounts |
| `src/modules/game/matches.service.ts` | Auto-increments `totalGames` in `recordMatch` and handles dynamic leaderboard sorting |
| `src/modules/users/users.controller.ts` | Updated validation to allow sorting by `totalGames` via query params |
| `src/services/userService.ts` (Frontend) | Fixed casing (`totalGames`) and updated the response interface |
| `src/pages/LeaderBoard.tsx` (Frontend) | Added sort button and dynamic stat display (Wins / XP / Games) |

---

## What was done

The main goal was to provide a clear view of player activity while ensuring optimal leaderboard performance.

### 1. Data persistence — `totalGames`

- **Auto-increment:** In `matches.service.ts`, every match conclusion now triggers an `increment: 1` on `totalGames` for both participants inside the Prisma transaction, ensuring data integrity even on draws.

### 2. Leaderboard evolution

- **Multi-criteria sorting:** The backend now supports three sorting axes — experience (`xp`), victories (`wins`), and total activity (`totalGames`).
- **Dynamic mapping:** The match service now returns the full stats block so the frontend can display contextual data per active filter.

### 3. Frontend / backend alignment

- **Casing fix:** Strict synchronization of sort keys between the API and the client (`totalgames` → `totalGames`).
- **Adaptive UI:** The React component adjusts its labels and displayed values using nested ternaries based on the active filter.

---

## Why

- **SQL performance:** Sorting by `totalGames` is now native and instant in the database — no joins or runtime aggregations needed.
- **User engagement:** Highlights the most active players, even those without the best win ratio.

---

### Setup

Reset your environment to apply the new schema and updated seed:

```sh
make nuke
make setup
make up
```

---

### Verify the seed

Check that dummy accounts have their stats correctly initialized:

```
http://localhost:1024/api/users/leaderboard?sortBy=totalGames
```
```
http://localhost:1024/leaderboard
```

All seeded users should appear with a non-zero `totalGames` value reflecting their match history.

---

### Test match recording

1. Connect two accounts and play a match.
2. After the match ends, verify that `totalGames` incremented by 1 for **both** players:

```
http://localhost:1024/leaderboard
```

3. Switch between the three leaderboard filters in the React UI and confirm the display updates correctly:

| Filter | Displayed stat |
|--------|---------------|
| `xp` | Experience points |
| `wins` | Total victories |
| `totalGames` | Total games played |

### DIFF
--- a/backend/prisma/schema.prisma
+++ b/backend/prisma/schema.prisma
@@ -23,7 +23,6 @@ model User {
   draws               Int            @default(0)
   xp                  Int            @default(0)
   totalGames          Int            @default(0)
-  winrate             Float          @default(0)
   isTwoFactorEnabled  Boolean        @default(false)
   twoFactorSecret     String?
   twoFactorTempSecret String?
--- a/backend/prisma/schema.prisma
+++ b/backend/prisma/schema.prisma
@@ -22,6 +22,8 @@ model User {
   losses              Int            @default(0)
   draws               Int            @default(0)
   xp                  Int            @default(0)
+  totalGames          Int            @default(0)
+  winrate             Float          @default(0)
   isTwoFactorEnabled  Boolean        @default(false)
   twoFactorSecret     String?
   twoFactorTempSecret String?
--- a/backend/prisma/seed.ts
+++ b/backend/prisma/seed.ts
@@ -112,7 +112,7 @@ async function main() {
 
     // ── XP & Update User ──────────────────────
     const xp = wins * 20 + draws * 10 + losses * 5;
-    await prisma.user.update({ where: { id: user.id }, data: { wins, losses, draws, xp } });
+    await prisma.user.update({ where: { id: user.id }, data: { wins, losses, draws, xp, totalGames: total } });
   }
   console.log('\x1b[32m✓ Updated Users\' XP and Stats\x1b[0m');
 }
--- a/backend/src/modules/game/matches.service.ts
+++ b/backend/src/modules/game/matches.service.ts
@@ -60,6 +60,7 @@ export class MatchesService {
             data: {
               wins: { increment: 1 },
               xp: { increment: 20 },
+              totalGames: { increment: 1 },
             },
           });
 
@@ -68,6 +69,7 @@ export class MatchesService {
             data: {
               losses: { increment: 1 },
               xp: { increment: 5 },
+              totalGames: { increment: 1 },
             },
           });
         } else if (result.winnerId === result.player2Id) {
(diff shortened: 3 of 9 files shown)


simo1616:
Thanks for the work on this!

I checked the current code and I think `totalGames` is not fully wired yet.

From what I can see:
- `recordMatch` still increments `wins`, `losses`, `draws` and `xp`, but I don’t see `totalGames` incremented for both players.
- `getGameLeaderboard` only accepts `'wins' | 'xp'`, so `totalGames` is not supported there yet.
- In `UsersController`, `safeSortBy` only keeps `xp`; every other value becomes `wins`. So `/api/users/leaderboard?sortBy=totalGames` currently falls back to wins.
- The leaderboard response only returns `id`, `username`, `xp` and `wins`, so the frontend cannot display `totalGames`.
- I also don’t see `totalGames` on the `User` model in the current schema, so maybe the Prisma schema/migration part is missing from the branch.

Can you check this before merge? The idea is good, but right now the API does not seem to support `totalGames` correctly yet.

simo1616:
My bad, I made a mistake in my previous review.

I probably mixed up files or looked at an older state of the branch. After re-checking the current code:
- `recordMatch` does increment `totalGames` for both players.
- `getGameLeaderboard` does accept `wins`, `xp`, and `totalGames`.
- `UsersController` does allow `totalGames` as a valid sort option.
- `/api/users/leaderboard?sortBy=totalGames` does not fall back to wins anymore.
- The leaderboard response does include `totalGames`.
- The `User` model does have a `totalGames` field.

So my previous comments about `totalGames` not being wired correctly are not valid. Sorry for the confusion.

simo1616 (approved):
Approved from my side.

I re-checked the current code after my previous mistake, and the `totalGames` flow now looks correctly wired across the schema, match recording, leaderboard sorting, API response, and frontend display.

Thanks for the clarification and sorry again for the confusion.

PR #250, opened 2026-05-03
[BACKEND][USERS] Validate pagination and add username search

### LINKED ISSUE #248: [BACKEND][USERS] Add username search to users endpoint
## Goal
Allow filtering users by username through `GET /users?search=...`.

## Tasks
- [ ] Add optional `search` query parameter
- [ ] Validate `search` as a string
- [ ] Add Prisma username filter
- [ ] Keep pagination working
- [ ] Test search + pagination together

## Constraints
- Do not expose private user fields
- Do not change response shape
- Do not touch frontend yet

Nico-Ry:
## Summary

This PR improves `GET /users` by adding validated pagination query parameters and an optional username search.

It also ensures cleaner input handling by rejecting empty search strings and trimming user input.

Closes #245  
Closes #248

---

## What was done

- Added `GetUsersQueryDto`
- Validated query parameters using `class-validator`:
  - `limit` (optional, integer, >= 0)
  - `offset` (optional, integer, >= 0)
  - `search` (optional string, min length 1, max length 50)
- Converted query params using `@Type(() => Number)`
- Kept max limit enforcement (`100`) inside the service layer
- Added optional `search` query parameter
- Implemented case-insensitive username search using Prisma
- Trimmed search input in the service to avoid issues with leading/trailing spaces
- Added Swagger documentation for all query parameters

---

## Reasoning

- DTO validation ensures:
  - type safety
  - rejection of invalid inputs (negative numbers, non-numbers, empty strings)
- Service layer handles:
  - business rules (default values, max limit cap)
- Search is:
  - case-insensitive
  - partial match (`contains`)
  - sanitized (trimmed input)

---

### Pagination

- `GET /api/users`
- `GET /api/users?limit=2`
- `GET /api/users?limit=2&offset=2`

### Search

- `GET /api/users?search=dummy`
- `GET /api/users?search=DUMMY`
- `GET /api/users?search=dummy&limit=2&offset=2`

### Validation

- `GET /api/users?limit=-5` → 400 ❌
- `GET /api/users?limit=abc` → 400 ❌
- `GET /api/users?search=` → 400 ❌ (min length 1)
- `GET /api/users?search=          ` → 400 ❌ (trim and then min length 1)
- `GET /api/users?search=     username` → handled correctly (trimmed input)
- `GET /api/users?search=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` (50 chars) → valid ✅
- `GET /api/users?search=aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa` (51 chars) → 400 ❌

### Swagger

- Verified `limit`, `offset`, and `search` appear in Swagger UI
- Verified descriptions and examples are correctly displayed

---

## Notes

- Pagination cap (`100`) is enforced in the service, not the DTO, to separate validation from business logic
- Empty search queries are rejected to avoid ambiguous behavior
- Search input is trimmed to ensure consistent filtering

---

## Future work

- Add sorting in a separate PR if needed (e.g. `sortBy=username|xp|wins` and `order=asc|desc`)
- Add advanced filters later only if the frontend needs them
- Integrate search into frontend later for friends, chat, or invites

For sorting later , the clean API would be :
- GET /users?search=nico&sortBy=username&order=asc
- GET /users?sortBy=xp&order=desc

### DIFF
--- a/backend/src/users/dto/get-users-query.dto.ts
+++ b/backend/src/users/dto/get-users-query.dto.ts
@@ -6,7 +6,6 @@ export class GetUsersQueryDto {
 	@Type(() => Number)
 	@IsInt()
 	@Min(0)
-	@Max(100)
 	limit?: number;
 
 	@IsOptional()
--- a/backend/src/users/dto/get-users-query.dto.ts
+++ b/backend/src/users/dto/get-users-query.dto.ts
@@ -1,4 +1,4 @@
-import { IsOptional, IsInt, Min, IsString, MaxLength } from 'class-validator';
+import { IsOptional, IsInt, Min, IsString, MaxLength, MinLength } from 'class-validator';
 import { Type } from 'class-transformer';
 import { ApiPropertyOptional } from '@nestjs/swagger';
 
@@ -26,10 +26,12 @@ export class GetUsersQueryDto {
 	@ApiPropertyOptional({
 		example: 'nico',
 		description: 'Optional case-insensitive username search.',
+		minLength: 1,
 		maxLength: 50,
 	})
 	@IsOptional()
 	@IsString()
+	@MinLength(1)
 	@MaxLength(50)
 	search?: string;
 }
--- a/backend/src/users/dto/get-users-query.dto.ts
+++ b/backend/src/users/dto/get-users-query.dto.ts
@@ -1,0 +1,17 @@
+import { IsOptional, IsInt, Min, Max } from 'class-validator';
+import { Type } from 'class-transformer';
+
+export class GetUsersQueryDto {
+	@IsOptional()
+	@Type(() => Number)
+	@IsInt()
+	@Min(0)
+	@Max(100)
+	limit?: number;
+
+	@IsOptional()
+	@Type(() => Number)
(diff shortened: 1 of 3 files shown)


simo1616 (approved):
Approved, this looks good to me.

I checked the PR and the implementation looks aligned with the description.

What I validated:
- `GET /users` now uses a `GetUsersQueryDto`.
- `limit` and `offset` are converted to numbers with `@Type(() => Number)`.
- invalid values like `limit=-5` or `limit=abc` are rejected.
- `search` has min/max length validation.
- empty search values are rejected.
- search values with only spaces are also handled correctly, because the service trims the value and throws a `BadRequestException` if it becomes empty.
- the service still caps the max limit to `100`.
- the Prisma search uses `contains` with `mode: 'insensitive'`.
- the user query uses `select`, so it does not load the full user object.
- Swagger docs are present with `@ApiPropertyOptional`.

For me this PR is good to merge.

PR #253, opened 2026-05-03
[BACKEND][USERS] Add validated avatar upload

### LINKED ISSUE #252: [BACKEND][USERS] Add validated avatar upload
## Goal
Allow authenticated users to upload a validated avatar image.

## Problem
The current profile update only accepts an avatar string. There is no backend-controlled upload flow, no file validation, and no safe ownership check for avatar changes.

## Tasks
- [x] Decide avatar storage approach
- [x] Add avatar upload endpoint
- [x] Accept multipart/form-data image upload
- [x] Validate image type
- [x] Validate image size
- [x] Convert image to safe DB representation
- [x] Update only the logged-in user's avatar
- [x] Return public user response
- [x] Document endpoint in Swagger
- [x] Test with Swagger/curl/browser

## Constraints
- No cloud storage
- Do not mix with email update
- Do not expose private user fields
- Do not allow users to update another user's avatar
- Keep existing public user response shape if possible

## Testing
- Upload valid PNG
- Upload valid JPEG
- Upload valid WebP if supported
- Reject non-image file
- Reject too-large file
- Confirm avatar changes in `/auth/me`
- Confirm avatar changes in `/users/:id`
- Confirm logout/auth behavior unaffected

Nico-Ry:
### Summary

This PR introduces a secure avatar upload flow for authenticated users and integrates it into the Profile UI.

Users can now upload avatar images directly from the frontend, with strong backend validation and protection against common file upload vulnerabilities.

---

### Backend — Avatar upload endpoint

```txt
POST /api/users/me/avatar
```

- Uses `multipart/form-data`
- Accepts a file under `avatar`
- Uses NestJS `FileInterceptor` (Multer)

---

#### File presence

```txt
Missing file → 400 Bad Request
```

#### File size (2 layers)

1. **Multer layer (prevents memory exhaustion)**

```ts
FileInterceptor('avatar', {
	limits: { fileSize: 200 * 1024 }
})
```

Expected result:

```txt
413 Payload Too Large
```

2. **Service layer (defense-in-depth)**

The service still checks the file size as a second safety layer in case interceptor configuration changes later.

---

#### File type (secure)

- Uses `fileTypeFromBuffer` instead of `file.mimetype`
- Prevents MIME spoofing attacks where a user lies about the uploaded file's `Content-Type`

Allowed:

```txt
image/png
image/jpeg
image/webp
```

Rejected:

```txt
400 Bad Request
Avatar must be a PNG, JPEG, or WebP image
```

---

### Backend — Security

- No `:id` in route, which prevents IDOR
- Uses `req.user.sub` only
- Prevents updating another user’s avatar
- Prevents oversized uploads before they reach the service
- Prevents MIME spoofing
- Handles Multer errors with a controlled 4xx response
- Adds a user existence check before updating, avoiding Prisma `P2025` becoming a 500
- Reuses `publicUserSelect` so private fields are not returned

---

### Backend — Storage

Avatar images are stored as:

```txt
data:<mime-type>;base64,<encoded>
```

Reason:

- No cloud storage
- No file system dependency
- No Docker volume needed
- Simple frontend usage with ``

---

### Backend — Constants

Moved avatar configuration to:

```txt
backend/src/users/avatar.constants.ts
```

Constants:

```ts
AVATAR_MAX_FILE_SIZE
AVATAR_ALLOWED_MIME_TYPES
```

This keeps validation rules centralized and avoids duplicating the max size or MIME list across the codebase.

---

### Backend — Error handling

Added:

```txt
backend/src/users/avatar-upload.exception-filter.ts
```

The exception filter handles upload errors such as:

```txt
LIMIT_FILE_SIZE → 413 Payload Too Large
```

This prevents upload errors from becoming uncontrolled 500 responses.

---

## Frontend — Avatar Upload UI

Added avatar upload support to the Profile page.

Files updated:

```txt
frontend/src/services/userService.ts
frontend/src/pages/Profile.tsx
```

### Behavior

- Upload button appears only in **Edit mode**
- "Change avatar" opens the file picker
- Frontend sends the selected file using `FormData`
- Backend validates and stores the avatar
- Avatar updates immediately on success

### Frontend flow

```txt
User selects file
→ Profile.tsx receives File
→ userService.uploadAvatar(file)
→ FormData appends field "avatar"
→ POST /api/users/me/avatar
→ backend validates and stores image
→ backend returns public user
→ React state updates avatar
```

Important field-name match:

```ts
formData.append("avatar", file);
```

matches the backend:

```ts
FileInterceptor("avatar")
```

---

### Test image setup

These images are only for local testing and are not committed to the repository.

```bash
mkdir -p test-images

curl -o test-images/big-avatar.png https://i.pravatar.cc/9950\?img\=10

curl -o test-images/avatar1.png https://i.pravatar.cc/150\?img\=1
curl -o test-images/avatar2.png https://i.pravatar.cc/150\?img\=2
curl -o test-images/avatar3.png https://i.pravatar.cc/150\?img\=3
curl -o test-images/avatar4.png https://i.pravatar.cc/150\?img\=4
curl -o test-images/avatar5.png https://i.pravatar.cc/150\?img\=5
curl -o test-images/avatar6.png https://i.pravatar.cc/150\?img\=6
curl -o test-images/avatar7.png https://i.pravatar.cc/150\?img\=7
curl -o test-images/avatar8.png https://i.pravatar.cc/150\?img\=8
curl -o test-images/avatar9.png https://i.pravatar.cc/150\?img\=9

echo "hello" > test-images/not-image.txt
```

---

### Authentication setup

Register a user:

```bash
curl -i \
	-d '{"username":"testuser","email":"test@example.com","password":"Secret123"}' \
	http://localhost:1024/api/auth/register
```

Login and store cookies:

```bash
curl -i \
	-c cookies.txt \
	-d '{"email":"test@example.com","password":"Secret123"}' \
	http://localhost:1024/api/auth/login
```

---

### Valid upload

```bash
curl -i \
	-b cookies.txt \
	-F "avatar=@test-images/avatar3.png" \
	http://localhost:1024/api/users/me/avatar
```

Expected:

```txt
201 Created
avatar starts with data:image/...
```

---

### Missing file

```bash
curl -i \
	-b cookies.txt \
	-X POST \
	http://localhost:1024/api/users/me/avatar
```

Expected:

```txt
400 Bad Request
Avatar file is required
```

---

### Invalid file type

```bash
curl -i \
	-b cookies.txt \
	-F "avatar=@test-images/not-image.txt" \
	http://localhost:1024/api/users/me/avatar
```

Expected:

```txt
400 Bad Request
Avatar must be a PNG, JPEG, or WebP image
```

---

### MIME spoofing exploit test

Create a fake HTML file:

```bash
echo '<html><body><h1>Not an image.</h1><script>alert("XSS: avatar bypass")</script></body></html>' > /tmp/exploit.html
```

Try to upload it while lying about the type:

```bash
curl -i \
	-b cookies.txt \
	-F "avatar=@/tmp/exploit.html;type=image/png" \
	http://localhost:1024/api/users/me/avatar
```

Expected:

```txt
400 Bad Request
Avatar must be a PNG, JPEG, or WebP image
```

This confirms that validation uses real file bytes, not the user-controlled multipart `Content-Type`.

---

### Too large file

```bash
curl -i \
	-b cookies.txt \
	-F "avatar=@test-images/big-avatar.png" \
	http://localhost:1024/api/users/me/avatar
```

Expected:

```txt
413 Payload Too Large
```

---

### Auth check

```bash
curl -i \
	-F "avatar=@test-images/avatar3.png" \
	http://localhost:1024/api/users/me/avatar
```

Expected:

```txt
401 Unauthorized
Missing authentication token
```

---

### Verify persistence

```txt
GET /api/auth/me
GET /api/users/:id
```

Expected:

- Avatar is updated
- `/users/:id` does not expose private fields
- No `passwordHash`
- No `twoFactorSecret`
- No `twoFactorTempSecret`

---

### Swagger test

In Swagger UI:

```txt
POST /api/users/me/avatar
```

Steps:

1. Log in first so the browser has the `access_token` cookie
2. Open `POST /api/users/me/avatar`
3. Click “Choose File”
4. Select a PNG, JPEG, or WebP image under 200 KB
5. Execute the request
6. Refresh the Profile page and confirm the avatar changed

---

### Frontend test

Steps:

```txt
Profile → Edit → Change avatar → select image
```

Expected:

- File picker opens when clicking the button
- Valid avatar uploads successfully
- Avatar updates immediately
- Invalid files show backend validation error through toast
- Too-large files show backend validation error through toast

---

### File size memory risk

Addressed by adding `limits.fileSize` directly to `FileInterceptor`.

This prevents large files from being fully processed by the service before being rejected.

---

### Multer error handling

Addressed by adding an upload exception filter.

Large files now return a controlled `413 Payload Too Large` response instead of causing an uncontrolled server error.

---

### Constants cleanup

Moved avatar size and MIME type configuration into module-level constants.

---

### MIME spoofing

Addressed by replacing `file.mimetype` validation with `fileTypeFromBuffer(file.buffer)`.

The backend now validates the actual file bytes.

---

### Missing user / Prisma P2025

Added a `findUnique` check before updating the avatar.

If the user no longer exists, the endpoint returns:

```txt
404 Not Found
User not found
```

instead of allowing Prisma `P2025` to become a 500.

---

## Notes

- Images are intentionally **not stored as files or in cloud storage**
- Base64 storage was chosen for simplicity and project constraints
- File size is strictly limited to avoid large DB payloads
- Backend validation remains the source of truth
- Frontend validation is only a usability layer, not a security layer

---

## Limitations / Future work

- Add image resizing/compression before storage
- Add avatar deletion/reset to default
- Consider local file storage with persistent Docker volume for larger avatars
- Consider object storage/CDN only if project constraints change
- Improve i18n for the new frontend upload button text

---

Closes #252


backend/src/users/users.controller.ts
```diff
@@ -10,52 +10,81 @@ import {
...
+				avatar: {
+					type: 'string',
+					format: 'binary',
+				},
+			},
+			required: ['avatar'],
+		},
+	})
+	@Post('me/avatar')
+	@UseInterceptors(FileInterceptor('avatar'))
```
Mikan95: The size check at `users.service.ts:122` runs after FileInterceptor has already loaded the entire file into memory. If a user flooded the upload with many large upload, it would exhaust the heap before any size check fire.

**users.service.ts:122**

```TypeScript
if (file.size > maxFileSize) {
	throw new BadRequestException('Avatar must be smaller than 200 KB');
}
```

**Suggested Fix:**

_Add the limits option to_ `FileInterceptor`

```TypeScript
@UseInterceptors(FileInterceptor('avatar', { limits: { fileSize: 200 * 1024 } }))
```

## After the Fix
Once limits are added Multer will throw a `MulterError` with code `LIMIT_FILE_SIZE` when a file does not respect the limits (i.e. it's too big).

NestJS does not automatically handle this and instead will throw a 5xx error instead of 4xx. 

Therefore we will need to add an exception filter either in the controller or globally:

```TypeScript
if (error instanceof MulterError && error.code === 'LIMIT_FILE_SIZE') {
    throw new BadRequestException('Avatar must be smaller than 200 KB');
}
```

backend/src/users/users.service.ts
```diff
@@ -107,4 +107,32 @@ export class UsersService {
...
+	
+		async updateAvatar(userId: number, file: Express.Multer.File | undefined) {
+		const allowedMimeTypes = ['image/png', 'image/jpeg', 'image/webp'];
+		const maxFileSize = 200 * 1024;
+
+		if (!file) {
+			throw new BadRequestException('Avatar file is required');
+		}
+
+		if (!allowedMimeTypes.includes(file.mimetype)) {
```
Mikan95: ## Potential Issue

This validates `file.mimetype`, but _Multer_ sets `mimetype` from the Content-Type header of the multipart body part,  not from the actual file bytes.

This means a user could send as curl command with a malicious file by lying about the Content-Type header.

**Example:**

```bash
curl -b cookies.txt \     
-F "avatar=@malicious.html;type=image/png" \
http://localhost:1024/api/users/me/avatar
```

## What is Multer?
**Multer** is an Express middleware for handling `multipart/form-data` (the encoding format for uploads via `HTML forms` or `curl -F`).  

When a request comes in, **Multer** intercepts it, parses the multipart body, and hands each field/file to the handler. 

> In NestJS, `FileInterceptor('avatar')` wraps **Multer** under the hood.
                                                                                                                                                                   
By default, **Multer** uses memory storage. It reads the entire file into a Buffer on `req.file.buffer` before the handler code runs. That's what `file.buffer.toString('base64')` is accessing. 

The `file.mimetype` field is populated **directly from the Content-Type header** of that multipart. (**Multer** never actually inspects the bytes)

## Testing the Exploit

**1. Create "Malicious" HTML File:**

```bash
echo '<html><body><h1>Not an image.</h1><script>alert("XSS: avatar bypass")</script></body></html>' > /tmp/exploit.html
```

**2. Login as Dummy5:**

```bash
curl -i -c /tmp/cookies.txt -H "Content-Type: application/json" -d '{"email":"dummy5@mail.com","password":"Dummy123"}' http://localhost:1024/api/auth/login
```

**3. Upload HTML file as Avatar (spoofed png):**

```bash
curl -i -b /tmp/cookies.txt -F "avatar=@/tmp/exploit.html;type=image/png" http://localhost:1024/api/users/me/avatar
```

> The `type=image/png` part does the "spoofing". It essentially disguises the HTML file as an image/png file.  
**>Expect:** `201 Created`

**4. Check what is stored in the avatar field:**

```bash
curl -s -b /tmp/cookies.txt http://localhost:1024/api/auth/me
```

> "avatar" should look something like: `data:image/png;base64,PGh0bWw+...` (long string of characters)

**5. Decode the stored base64 file:**

```bash
curl -s -b /tmp/cookies.txt http://localhost:1024/api/auth/me | python3 -c "
import sys, json, base64
data = json.load(sys.stdin)
avatar = data.get('avatar', '')
if ',' in avatar:
    decoded = base64.b64decode(avatar.split(',')[1]).decode()
    print('\033[1;91mDecoded avatar content:\033[0m')
    print(decoded)
else:
    print('\033[1;92mNo avatar or unexpected format:\033[0m', avatar)
"
```

> **Note:** if you don't have Python 3 installed on your system this command may not work.

backend/src/users/users.service.ts
```diff
@@ -107,4 +107,32 @@ export class UsersService {
 			throw new BadRequestException('Failed to update user');
 		}
 	}
+	
+		async updateAvatar(userId: number, file: Express.Multer.File | undefined) {
```
Mikan95: Small detail but this is indented one increment too far! 👍 (2 tabs instead of 1)

backend/src/users/users.service.ts
```diff
@@ -107,4 +107,32 @@ export class UsersService {
...
+			throw new BadRequestException('Avatar must be a PNG, JPEG, or WebP image');
+		}
+
+		if (file.size > maxFileSize) {
+			throw new BadRequestException('Avatar must be smaller than 200 KB');
+		}
+
+		const avatarDataUrl = `data:${file.mimetype};base64,${file.buffer.toString('base64')}`;
+
+		const updatedUser = await this.prisma.user.update({
```
Mikan95, line 144: ## Minor Suggested Polish

I think like for `updateUser` we should add a `findUnique` check at the beginning of this function. To avoid a scenario where the user is deleted between JWT issue and the request. (Yes, quite unlikely, but as `updateUser` already does this, why not add it to `updateAvatar` as well 👍 ).


## Reasoning
`prisma.user.update` would throw `PrismaClientKnownRequestError` with code `P2025` if the user had somehow been deleted in between the issue of its JWT token and the request to upload a file. 
> Essentially, it's prisma saying "I went to update row x but nothing was there"

NestJS would in turn produce an `500 Internal Server Error` which does not provide any useful information to the user.

We agree that this is an unlikely scenario and that this is more of a "make it consistent with `updateUser`" and "professional polish" kind of change.

backend/src/users/users.service.ts
```diff
@@ -107,4 +107,32 @@ export class UsersService {
 			throw new BadRequestException('Failed to update user');
 		}
 	}
+	
+		async updateAvatar(userId: number, file: Express.Multer.File | undefined) {
+		const allowedMimeTypes = ['image/png', 'image/jpeg', 'image/webp'];
+		const maxFileSize = 200 * 1024;
```
Mikan95: Another Minor Detail. But it may be worth moving these outside the `updateAvatar` function so they act as a module-level constant rather than inline variables.

I know at the moment we don't have any other avatar-related methods, but if we did add some for some reason. Then having them inline, would mean nedding to duplicate them and update them in two different places if we decide to change `MAX_AVATAR_SIZE` for example.

Mikan95 (changes requested):
## What's Done Well                                                                                                                                                 
                                                                                                                                                                   
- Correct IDOR prevention: route has no :id param, only `req.user.sub` is used, meaning a user cannot update another user's avatar.
- `publicUserSelect` reuse ensures no private fields leak in the response.
- Auth is covered globally; no @Public() risk.                                                                                                                   
- Swagger annotations are complete and correct.                                                                                                                  
- Test commands in the PR description are comprehensive and well-structured.

## A couple of Issues to Handle

| Severity of change | Issue | Potential Risk |
| --- | --- | --- |
| Needed | MIME type check reads request header, not file magic bytes | possibility of spoofing |
| Needed | No `fileSize` limit on `FileInterceptor` | Full file uploaded by user is loaded into memory before a size check occurs |
| Needed | `MulterError` for size overflow is unhandled | Returns 500 instead of `413 Content Too Large` |
| Medium-low | `prisma.user.update` in `updateAvatar` has no guard for P2025 | Returns 500 on missing user |
| low | A couple of other small polishes and changes listed in my comments! | Code quality degradation |

Mikan95 (approved):
Great work on the changes! I like the addition of the upload button when editing the profile! It's a nice touch!

👍

PR #255, opened 2026-05-04
[INFRA] - Split .env into .env.dev and .env.prod

### LINKED ISSUE #235: [INFRA] - Split .env into .env.dev and .env.prod
## Description

The project currently uses a single `.env` file shared by both the dev and prod stacks.
Switching between stacks requires re-running `make setup` or manually editing variables —
particularly `DOMAIN` and all URL-based variables that differ between dev (`http`, port `1024`)
and prod (`https`, port `8443`). This creates a fragile, stateful workflow where it is easy
to run the wrong stack with the wrong config.

## Goal

`make setup` generates both `.env.dev` and `.env.prod` in a single run.
Each stack consumes only its own file via `--env-file` in the corresponding Makefile targets.
Both files can exist simultaneously, with their own `DOMAIN` and protocol-correct URLs,
allowing dev and prod to be configured and started independently without any reconfiguration.

## Scope

- [ ] Frontend
- [ ] Backend
- [ ] Database
- [x] Docker / Infrastructure
- [ ] Documentation
- [ ] Other

## Proposed Tasks

- [ ] Replace `.env.example` with `.env.dev.example` and `.env.prod.example`
- [ ] Rewrite `make setup` / `_setup-apply` in `make/setup.mk` to generate both files
- [ ] Update all `docker compose` invocations in `Makefile` to pass `--env-file`
- [ ] Update `-include .env` in `Makefile` to `-include .env.dev`
- [ ] Update `_check-required-files` to require both `.env.dev` and `.env.prod`
- [ ] Update `.gitignore` to track new example files
- [ ] Update `make swagger` to grep `DOMAIN` from `.env.dev` and `.env.prod` instead of `.env`
- [ ] Test locally
- [ ] Update documentation if needed

## Notes

This will be useful for the integration of OAuth + 2FA in `make setup`. By having two separate files I can have `make setup` simply write the correct URLs in both files (http vs https, 1024 vs 8443) instead of requiring devs to run separate `make setup` and `make p-setup` targets depending on which stack they want to run.
> The change to have OAuth and 2FA handled by `make setup` will follow this change.
> Think of this as a kind of pre-requisite.

The `swagger` Makefile target currently reads `$(DOMAIN)` from the included `.env`.
With two files I will need to update it to grep `DOMAIN` from each file separately.

Docker Compose variable substitution in compose files (e.g. `${POSTGRES_VERSION}`)
should continue to work as it reads from whichever file is passed via `--env-file`.

Mikan95:
Closes #235 

## Checklist

- [x] Target branch is `dev`

### **File Changes:**
- Deleted `.env.example`, replaced by `.env.dev.example` and `.env.prod.example`
- Updated `.gitignore` to track the two new example files instead of `.env.example`
- Removed dead `JWT_SECRET=${JWT_SECRET}` env var from `compose.dev.yaml` backend service

### **Make Setup:**
- Rewrote `make/setup.mk`: 
  - `_setup-apply` now generates both `.env.dev` and `.env.prod` from their respective examples
  - Added `_domain-dev` and `_domain-prod` targets for per-stack DOMAIN prompts
  - Removed `DEV_ONLY_ENV_VARS` (Obsolete as `BACKEND_PW`, `JWT_SECRET` are Docker secrets in both stacks)
  - Updated `REQUIRED_FILES` to require `.env.dev` and `.env.prod` instead of `.env`

### **Infra & Makefile:**

- Changed `Makefile` include from `-include .env` to `-include .env.dev`
- Added `ENV_DEV` and `ENV_PROD` variables and applied `--env-file` to every `docker compose` invocation in `Makefile` and `make/clean.mk`

### **Make Swagger:**
- Rewrote `swagger` target to grep `DOMAIN` from each env file independently

### **Behaviour Change:**

- `make setup` prompt now defaults to **yes** on empty input (`[Y/n]` instead of `[y/N]`) — pressing Enter proceeds with setup rather than cancelling it
- The **dev** and **prod** stacks now each run on different dotenv files. This allows devs to define different behaviour for each stack. 
  - **For example:**  
  Running **dev** on localhost (127.0.0.1) and **prod** on Local LAN (168.192.x.x).

## Why

A single `.env` shared between dev and prod required manually editing `DOMAIN` and re-running setup before every prod deployment. Splitting into `.env.dev` and `.env.prod` lets `make setup` configure both stacks in one pass and eliminates the risk of accidentally deploying with dev values. Each `docker compose` invocation now explicitly loads the correct file, removing reliance on Make's `export` to implicitly pass variables to Docker Compose.

### Prod Test 1 — Fresh setup and prod stack start

**1. Full wipe and setup:**

```bash
make nuke
make setup
```

> Answer `n` to both DOMAIN prompts to use `127.0.0.1` defaults.

**2. Confirm both env files were created and both DOMAIN are set to whatever you did during the setup prompt:**

```bash
ls .env.dev .env.prod
grep DOMAIN .env.dev .env.prod
```

**3. Start the prod stack:**

```bash
make prod
```

**4. Confirm the stack is healthy:**

```bash
make ps
```

**5. As an extra measure verify URLs in `.env.prod` are all `https` and use port `8443`:**

```bash
grep -E \
"FORTYTWO_REDIRECT_URI|FRONTEND_OAUTH_SUCCESS_URL|VITE_OAUTH_42_START_URL|TWO_FACTOR_URL" \
.env.prod
```

---

### Dev Test 1 — Dev stack after nuke

**1. Remove some Required files to 'simulate a nuke':**

```bash
rm .env.dev .env.prod
```

**2. build the dev stack:**

```bash
make up
```

> `_check-required-files` must trigger `_setup-apply` automatically. Both env files and all secrets must be created before the stack starts. (you will have to go through the prompting again).

**3. Confirm stack is healthy:**

```bash
make ps
```

**4. As an extra measure verify URLs in `.env.dev` are all `http` and use port `1024`:**

```bash
grep -E \
"FORTYTWO_REDIRECT_URI|FRONTEND_OAUTH_SUCCESS_URL|VITE_OAUTH_42_START_URL|TWO_FACTOR_URL" \
.env.dev
```

---

### Swagger Test — Swagger reads correct DOMAINs

**1. Run `_setup-apply` and set domains to different IPs:**

```bash
make _setup-apply
```

> Set dev to Local LAN and prod to localhost or vice versa

**2. Check both DOMAIN are different:**

```bash
Automatically setup DOMAIN? [Y/n]
[DEV]  Set DOMAIN to local LAN? [Y/n]
       Using custom DOMAIN: '192.168.1.67' <---- This should be different to PROD domain

[PROD] Set DOMAIN to local LAN? [Y/n] no
       Using default DOMAIN: '127.0.0.1' <---- This should be different to DEV domain
```
> Or run the command below:

```bash
grep DOMAIN .env.dev .env.prod
```

**3. Run swagger:**

```bash
make swagger
```

> Dev URL must match `DOMAIN` in `.env.dev`, prod URL must match `DOMAIN` in `.env.prod`.


Nico-Ry:
# PR Review — Environment Separation / Make Setup

Hi! I reviewed and tested this PR.

First of all, great work on the env separation. The idea of splitting `.env.dev` and `.env.prod` makes a lot of sense, especially because it removes the need to manually edit the same `.env` file when switching between dev and prod. I also like that every `docker compose` command now explicitly receives the correct env file with `--env-file`, because that makes the behaviour much clearer and less implicit.

I tested the main flows described in the PR and most of the setup works well, but I found a few issues that should be fixed before merging.

---

## What I tested

I tested the following parts:

- `make setup`
- creation of `.env.dev`
- creation of `.env.prod`
- creation of the secrets files
- different DOMAIN values for dev and prod
- `make up`
- `make prod`
- `make ps`
- `make swagger`
- `make clean`
- `make nuke`
- running `make nuke` more than once
- behaviour when required env files are missing
- behaviour when invalid input is entered during setup prompts

The main env separation works, and I was able to run both dev and prod stacks with separate env files.

I also confirmed that `make swagger` correctly reads the DOMAIN values from the separate env files:

```bash
make swagger
```

Example result:

```bash
═════ DEV ═════════════════════════
http://127.0.0.1:1024/api/api-docs

═════ PROD ════════════════════════
https://172.19.116.8:8443/api/api-docs
```

## Final review

The env separation is a strong improvement and the direction is correct.
I think this will be a very useful infra improvement for the team. Great job boss :dancing_men:

Nico-Ry (approved):

Mikan95:
Thank you for the comments! and Taking the time to test features in depth!
Regarding the issues you raised, most of the behaviour described is expected behaviour.

## Blocking issue 1 — make nuke fails when .env.dev is missing

Due to the nature of having two separate dotenv files, Docker and Docker compose need `.env.dev` and `.env.prod` to know which containers to delete.
With either of those files missing the compose commands that delete images/containers etc. will fail.

If we remove the `--env-file .env.dev` from the command, docker compose does not delete the images when runnning make nuke. (I tested this to see if there was a way to silence the error message).

Therefore, I think this is just expected behaviour and in most cases make nuke will be used to wipe the project and thus in cases where the dotenv files will be present.


## Blocking issue 2 — invalid setup prompt input can break setup state

More so than "breaking setup state" it leaves the state as the default example and allows users to edit the dotenv files manually.

Make setup exists for convenience and quick setup. But if users respond no to the setup then it is expected that they must edit the dotenv files manually to better fit there needs.


>For example, instead of stopping immediately and leaving a broken state, it could keep asking until the user enters >a valid answer:
>
>```bash
>Please answer y or n.
>Automatically setup DOMAIN? [Y/n]
>```

Regarding the above, once again this is expected behaviour. If the user refuses the automatic setup of domain. Then the process exits and leaves the user to update it manually. If you answer yes, then you run through the prompts for automatic DOMAIN setup.


## Issue 3 — make clean does not stop running containers

`make clean` does not stop running containers. It only removes dangling images, clears any build cache that it can and removes any containers related to the active stack that **are already stopped.**

I understand that last part is a little confusing. If I'm frank, this is a little obsolete and an artifact of the initial architecture. Before I gave specific names to our containers in the compose.yaml files, each time you ran `make up` Docker Compose would automatically untag the previous version of the container (if any existed) and create a new version of the container. This leed to old/obsolete containers taking up space. The logic behind removing only stopped containers was to ensure only those "old" versions of the containers were removed but kept the active or current ones.

`make clean` is intended as a low impact non-destructive clean. To free up as much space as possible without deleting anything valuable! :+1: 

---

I hope this clears things up a little! And if you have any other questions or anything is unclear please let me know :+1:

PR #262, opened 2026-05-05
Fix deploy.yml

simo1616:
Edited deploy command to specify the correct dotenv file. In this case: ```--env-file .env.prod```

This change was made as the dotenv file has now been split in two different files for each environment (prod vs dev)

view #255


Mikan95 (approved):
Perfect! This was a necessary change so the deployment Github Action works with the recent changes to dotenv files.

PR #264, opened 2026-05-06
[FRONT-BACK] Add basic live chat module

simo1616:
# Summary

This PR adds a simple global live chat using Socket.IO.

The chat is intentionally basic for now.
It is a live-only chat, so messages are not stored in the database and there is no history after refresh.

### Frontend
- Added `/chat` route on the frontend
- Added `Chat` page with authentication check (redirects if not logged in)
- Added `BasicChat` component using Socket.IO
- Socket instance managed with `useRef` to avoid re-connections on re-renders
- Sends `join_chat` event on socket connect
- Displays messages in format: `HH:MM @username: message content`
- Extracted `User` interface to shared `src/type/user.types.ts` (removed local duplicate)

### Backend
- Added separated `ChatModule` importing `AuthModule` and `UsersModule`
- Added separated `ChatGateway` with:
  - `@UseGuards(JwtAuthGuard)` for WebSocket authentication
  - `@UsePipes(ValidationPipe)` at gateway level (global pipes do not cover WebSockets)
- Added `SendChatMessageDto` with `class-validator`:
  - `@IsString()`
  - `@MinLength(1)` — rejects empty messages
  - `@MaxLength(500)` — rejects messages over 500 chars
- Modified `AuthModule` to also export `JwtModule` so sub-modules can import auth cleanly
- Backend fetches real username from DB via `UsersService.getUser()` — no hardcoded username
- Backend broadcasts messages to all connected users via `this.server.emit`

### Socket events
| Event | Direction | Description |
|---|---|---|
| `join_chat` | client → server | Client joins the chat |
| `chat_ready` | server → client | Server confirms client is ready |
| `chat_send` | client → server | Client sends a message |
| `chat_message` | server → all | Server broadcasts message to all |
| `chat_error` | server → client | Server rejects invalid message |

### What was NOT changed
- Game gateway logic untouched
- No DB schema modified
- No message history added

## Why

This is the first simple version of the chat.
The goal was to keep it small and safe, without mixing it with the game gateway or adding DB logic too early.

### Dev mode

```bash
make dev
```

### Prod mode

```bash
make prod
```

### Test steps

1. Open the app and login with two different users (two browsers or two sessions)

2. Go to `/chat` in both sessions

3. Send a message from user A

   Expected:
   - user A sees the message with format `HH:MM @username: message`
   - user B receives the message in real time with the real username

4. Send a message from user B

   Expected:
   - user A receives it in real time

5. Try to send an empty message

   Expected:
   - message is not sent
   - `chat_error` is emitted by the backend

6. Try to send a message longer than 500 chars

   Expected:
   - message is rejected
   - `chat_error` is emitted by the backend

7. Refresh the page

   Expected:
   - old messages are gone (normal — no persistence)

8. Test the game

   Expected:
   - game still works normally
   - existing game sockets were not changed

## Notes

This PR does not include:
- Private chat
- Friends system
- Profile integration
- DB persistence
- Chat history
- Final UI design
- Notifications

Those can be added later in another PR if needed.

Closes #257
Closes #258
Closes #259


nasdhn (approved):
## ✅ PR Review — Approved

Tested all scenarios from the description manually using two separate browser sessions (Chrome + Safari) logged in with two different accounts:

- Real-time messaging works correctly in both directions
- Message format `HH:MM @username: message` displays as expected
- Username is correctly fetched from DB server-side — no spoofing possible from the client
- Empty message blocked, `chat_error` emitted by the backend 
- Message over 500 chars blocked, `chat_error` emitted by the backend 
- Page refresh correctly clears the chat (no persistence — expected behavior) 
- Navigated to the game after testing the chat — game gateway is fully intact, no regression 

On top of the requested tests, I also ran a **spam test** — sending messages rapidly and repeatedly from both sessions at the same time. The broadcast held up correctly with no duplicates, dropped messages, or race conditions observed.

---

**One UI bug spotted** — not a blocker for this PR since the final design is explicitly out of scope, but worth tracking before the UI pass:

When the chat fills up with a large number of messages, the message list overflows outside its container and continues rendering on a plain white background outside the page layout. Flagging it now so it does not get lost.

PR #271, opened 2026-05-10
DOC[GAME]: Swagger API documentation + chat & home fixes

simo1616:
## Summary

Swagger documentation for the game module REST API, plus related frontend fixes for the chat and home page.

---

### `game.controller.ts`
- Added `@ApiTags('Game')` on the controller
- Added `@ApiCookieAuth()` on protected routes (`POST /game/create`, `GET /game/active`)
- Added `@ApiOperation`, `@ApiResponse`, and `@ApiParam` on all 5 endpoints
- Documented success responses (200/201) and error responses (401, 404)

### `dto/play-move.dto.ts`
- Added `@ApiProperty` on `gameId`, `r`, `c` fields with examples, descriptions, and min/max values

### `dto/game-result.dto.ts`
- Added `@ApiProperty` on all required fields (`player1Id`, `player2Id`, `scoresP1`, `scoresP2`)
- Added `@ApiPropertyOptional` with `nullable: true` and enum on `winnerId` and `endReason`

---

### `Home.tsx`
- Replaced manual `fetch('/api/game/create')` with `gameApi.createGame()` from the centralized API service
- Added `createError` state to display a user-facing error message when game creation fails
- Removed unused `console.log`

### `BasicChat.tsx`
- Fixed layout: messages zone is now scrollable (`flex-1 overflow-y-auto`), input stays at the bottom
- Added close button (✕) via `onClose` prop — replaces the floating "close chat" button
- Fixed send button scroll bug: added `hover:scale-100` to prevent layout overflow on hover

### `Chatbar.tsx`
- Now accepts and forwards `onClose` prop to `BasicChat`

### `Dashboard.tsx`
- "Open chat" button now only visible when chat is closed (`!isChat`)
- Passes `onClose={handleChatClick}` to `Chatbar`

---

## Checklist
- [x] Swagger UI tested at `/api/api-docs`
- [x] All game endpoints documented with operations, responses, and params
- [x] DTOs have `@ApiProperty` / `@ApiPropertyOptional`
- [x] Protected routes have `@ApiCookieAuth()`
- [x] Frontend uses centralized API service instead of raw fetch
- [x] Chat close UX improved


Nico-Ry (approved):
Tested everything described in the PR and overall it works great, very nice addition to both the backend documentation and the frontend cleanup.

Swagger docs are much clearer now and the DTO documentation is properly exposed. I also tested the frontend fixes (chat behavior, close button flow, centralized API usage) and everything behaved correctly.

I only found 1 issue during testing:

`GET /game/:id/history`

When requesting the history of a game that was never actually started, the endpoint currently returns a `500 Internal Server Error`. I do not think this should be treated as a server error because the backend itself is still functioning correctly. This looks more like an invalid state / missing resource case, so a `4xx` response (`404` or possibly `400`) would probably be more appropriate than `500`.

Apart from that, everything looks good to me.

About the `@ApiCookieAuth()` addition:
 I confirmed that protected actions cannot be accessed while unauthenticated (game creation, active game access, etc.), so the protection is working correctly.

I also noticed the Swagger lock icon appears on `/game/create` and `/game/active`, which matches the routes where `@ApiCookieAuth()` was added.

Great work 👍

PR #284, opened 2026-05-11
[FIX] error handling, cleanup & 404 page

simo1616:
## Summary

This PR addresses review feedback and adds a small cleanup pass on the game module and frontend.

It improves error handling in the game history endpoint, removes unused/duplicate frontend code, adds a proper 404 page, and updates the app metadata/favicon.

---

### Bug fixes

- `GET /game/:id/history` now returns `400 Bad Request` instead of `500 Internal Server Error` when the game is not finished.
- Replaced generic `throw new Error(...)` with proper NestJS exceptions in `game.service.ts`:
  - `BadRequestException`
  - `ForbiddenException`

### Cleanup

- Removed an unreachable/dead `console.log` in `Game.tsx`.
- Removed duplicate `/leaderboard` route in `App.tsx`.

### Frontend features

- Added a `404 Not Found` page using the existing `EmptyStateCard` component.
- Added a wildcard route `*` in `App.tsx` to catch unknown URLs.

### Metadata

- Updated `index.html` with:
  - proper app title
  - improved description
  - favicon support using `.ico` and `.png`

---

## How to test

| Scenario | Expected result |
|---|---|
| Go to `/une-route-inexistante` | The 404 page is displayed |
| Call `GET /game/:id/history` on an unfinished game | Returns `400 Bad Request` instead of `500` |
| Check the browser tab | Title is `Ft_transcendence` and favicon is visible |
| Test `/game/:id/history` in Swagger with a valid ongoing game ID | Returns `400 Bad Request` |

---

## Notes

This PR does not introduce new game logic.  
It mainly improves stability, error handling, routing cleanup, and app metadata.


backend/src/modules/game/game.controller.ts
Mikan95, line 59: ## Missing `400` response in Swagger docs

The endpoint can now return three different HTTP status codes:

- `200` — game is finished, history is returned
- `404` — no game found with that ID
- `400` — game exists but is not finished yet

The `200` and `404` cases are already documented with `@ApiResponse`. The `400` case is not. This means the Swagger UI at `/api-docs` does not mention that a `400 Bad Request` is possible, which is misleading for anyone using the docs to understand this endpoint.

**Fix — add this decorator above the endpoint:**

```ts
@ApiResponse({ status: 400, description: 'Game is not finished yet' })
```

It should sit alongside the other `@ApiResponse` lines, like this:

```ts
@ApiResponse({ status: 200, description: 'Returns match history' })
@ApiResponse({ status: 400, description: 'Game is not finished yet' })
@ApiResponse({ status: 404, description: 'Game not found' })
@Get(':id/history')
```

Small change, but it keeps the documentation honest and makes the API easier to use.

Mikan95 (changes requested):
## Review Summary

Nice work on this one! The error handling changes are clean and correct — tested `GET /game/:id/history` on an active game and confirmed it now returns `400` instead of `500`. The 404 page works well too, unknown routes are handled gracefully. Using proper NestJS exceptions instead of generic `throw new Error` is a good habit and makes the API behave predictably.

Two small things to look at before we merge (see comments below).

---

## 1. Missing `400` response in Swagger docs (`game.controller.ts`)

See my comment in `backend/src/modules/game/game.controller.ts`

## 2. Unused and duplicate favicon files in `public/`

**Files:** `frontend/public/`

**What I noticed:**

10 favicon files were added, but `index.html` only references 5 of them. The other 5 are not linked anywhere, so the browser never loads them.

On top of that, 3 of those unreferenced files appear to be exact copies of files that are already referenced:

| File added | Looks identical to |
| --- | --- |
| `favicon-32x32.png` | `favicon-32.png` |
| `favicon-64.png` | `favicon.png` |
| `favicon-master-512.png` | `favicon-512.png` |

The two remaining unreferenced files (`favicon-48.png`, `favicon-512.png`) also don't appear to be used.

However, maybe I am missing something. 
If these files are needed for a specific reason (a PWA manifest, a future feature, a tool that expects them, etc.) that would make total sense. Please just leave a note so we know why they are there.

If they are not needed, it would be good to remove them before merge, since files added to git stay in the history even if deleted later.

Mikan95 (approved):
Great job! Thanks for adding the 400 response to the Swagger docs and cleaning up duplicate favicon images!

PR #285, opened 2026-05-12
[FRONTEND] : Frontend UI/UX Rework — Lobby, Spectate, History & Navigation

### LINKED ISSUE #279: [FRONTEND] : Add Draw Rule Explanation to Home Rules Card and Dedicated Rules Page
## Description

Update the game rules displayed in the Home page rules card and the dedicated Rules page to include the draw condition after more than 20 moves.

The goal is to ensure players clearly understand all game-ending conditions directly from the UI.

---

### Home page

- Update the rules card content
- Add the draw rule explanation alongside the existing game rules

### Dedicated Rules page

- Add the draw rule to the full rules description
- Keep formatting and spacing consistent with the current layout

### Draw rule

- A game ends in a draw after more than 20 moves without a winner

---

## Technical notes

- Update all i18n translations (EN / FR / ES)
- Preserve existing card/page layouts
- Keep wording consistent across all rule displays

---

## Goal

Improve gameplay clarity by documenting the draw condition everywhere the game rules are displayed.

ldsr18:
## Summary

This PR introduces a large frontend UI/UX rework focused on :

- lobby/game creation flow
- live spectate system
- empty states consistency
- sidebar/topbar navigation improvements
- profile/history polish
- responsive layouts
- full EN/FR/ES translations

The goal was to make the application more coherent visually while improving the user experience around games, navigation and authentication states.

---

# Main Features

### Added persistent lobby system

- create game flow with invite link generation
- waiting lobby persistence after refresh using `/game/active`
- automatic redirect when opponent joins
- polling system for active game detection
- cancel game support from lobby
- copy invite link support

### Available games improvements

- display waiting public games directly inside Play page
- display creator avatar and username
- exclude creator game from joinable games list
- loading states
- empty states

---

### Added live games page

- display all currently active games
- reusable shared card design
- avatars and usernames support
- watch game action
- loading states
- empty states

### Added translations

- EN / FR / ES support
- localized live games counter using i18n pluralization

---

### Reworked empty history state

- unified empty state design
- aligned layout with other pages/cards

---

### Refactored profile layout

- improved edit mode UI
- fixed avatar alignment during edit mode
- added profile navigation from topbar button

---

### Added draw rule

- draw after 20 moves
- integrated directly into rules card

### UI improvements

- fixed icon alignment inconsistencies across languages
- stabilized text layout using fixed icon containers
- improved spacing and readability

---

### Sidebar improvements

- moved language switcher from topbar to sidebar footer
- moved login/logout actions to sidebar footer (still can login from topbar for now)
- improved spacing and navigation consistency
- added spectate page entry
- added rules page entry

### Topbar improvements

- simplified authenticated user section
- clickable `ft_transcendance` title redirecting to home
- clickable user button redirecting to profile
- improved username/avatar alignment
- preserved login access for non-authenticated users

---

### Added/updated full translations

- English
- French
- Spanish

---

# UI/UX Improvements

- unified empty state system
- reusable card structure
- cleaner spacing and alignment
- stabilized scrolling areas
- improved consistency between pages

---

# Files Modified

## Frontend

- Play page
- Spectate page
- History page
- Profile page
- Sidebar
- Topbar
- GameRules
- EmptyStateCard
- GameHistoryCard
- translations EN/FR/ES

---

# How to Test

## Lobby System (Play page)

- create a game
- verify invite link generation
- refresh page and verify lobby persistence
- join from another account
- verify automatic redirect
- cancel created game

## Available Games

- verify avatars/usernames appear correctly
- verify creator cannot join own game (game should not even appear)
- verify loading and empty states

## Spectate

- start active games
- verify live games appear
- verify spectate button redirects correctly
- verify empty state when no games are active

## History

- verify history list rendering
- verify empty history state

## Profile

- edit profile
- verify avatar alignment

## Navigation

- click sidebar entries
- click `ft_transcendance` title => home
- click topbar user button => profile
- verify login/logout placement

## i18n

- switch between EN / FR / ES
- verify translations and pluralization

Closes #277 
Closes #278 
Closes #279 
Closes #280 
Closes #281


frontend/src/i18n/es.json
```diff
@@ -2,115 +2,148 @@
...
+			"cancel": "Cancelar partida"
+		},
+		"availGames": {
+    		"title": "Partidas disponibles",
+    		"subtitle": "Partidas públicas",
+    		"desc": "Únete a una sala en espera creada por otro jugador",
+    		"refresh": "Actualizar partidas",
+    		"empty": "No hay partidas disponibles",
+    		"waiting": "Esperando a un oponente",
+    		"join": "Unirse"
```
Mikan95, line 103: ### Spanish translations entirely broken (`es.json`)

The closing `}` for the new `"play"` block is missing. Because of this, everything from `"profile"` to the end of the file is nested **inside** `"play"` instead of being at the top level.

The actual structure in the file right now:

```json
"play": {
    "myGames": { ... },
    "availGames": { ... },
    "profile": { ... },       <- should NOT be inside play
    "leaderboard": { ... },   <- should NOT be inside play
    "friends": { ... },       <- should NOT be inside play
    "history": { ... },       <- etc.
    "buttons": { ... },
    "auth": { ... }
}
```

This means all lookups like `t("profile.loading")`, `t("leaderboard.title")`, `t("history.empty")`, etc. will fail silently in Spanish, the keys just don't exist at the right path. Everything still works in EN and FR because their `"play"` blocks are closed correctly.

**Fix:** add a closing `},` after the `"availGames"` block, before `"profile"` starts:

```json
    "play": {
        "myGames": { ... },
        "availGames": { ... }
    },

    "profile": { ... },
```

frontend/src/pages/Play.tsx
```diff
@@ -0,0 +1,300 @@
...
+      			method: "POST",
+    		});
+    		setCreatedGameId(null);
+   			await fetchGames();
+  		} catch (error) {
+    		console.error("Failed to cancel game:", error);
+  		}
+  	}
+
+	useEffect(() => {
```
Mikan95, line 83: ### Polling runs even when user is not logged in

This `useEffect` polls `/api/game/active` every 1500ms and is registered before the `if (!user)` guard that returns early. In React, hooks always run before a conditional return, so the interval is set up and starts making API calls even when the user is not logged in. Those calls return 401 and are silently caught. 
However, they still happen every 1.5 seconds for as long as the page is open.

**Fix:** add a user check inside the effect:

```ts
useEffect(() => {
    if (!user) return;
    // ...polling logic...
}, [navigate, user]);
```

frontend/src/pages/Play.tsx
```diff
@@ -0,0 +1,300 @@
...
+    	if (!createdGameId)
+			return;
+
+  		try {
+    		await fetch(`/api/game/${createdGameId}/leave`, {
+      			method: "POST",
+    		});
+    		setCreatedGameId(null);
+   			await fetchGames();
+  		} catch (error) {
```
Mikan95, line 78: ### Cancel game does not check if the request succeeded

`handleCreateGame` correctly checks `if (!response.ok)` before updating state. `handleCancelGame` does not. It calls `setCreatedGameId(null)` and `fetchGames()` regardless of whether the server returned an error. If the cancel request fails, the UI will show "no active game" when the game was not actually cancelled.

**Fix:** add the same response check as in `handleCreateGame`:

```ts
const response = await fetch(`/api/game/${createdGameId}/leave`, { method: "POST" });
if (!response.ok) throw new Error(`HTTP error ${response.status}`);
```

backend/src/modules/game/game.controller.ts
```diff
@@ -49,6 +54,11 @@ export class GameController {
     return this.gameService.getGameById(gameId);
   }
 
+  @Post(':id/leave')
```
Mikan95: ### New `POST /game/:id/leave` endpoint is missing all Swagger documentation

The new endpoint has no decorators at all, so it won't appear in Swagger docs and is inconsistent with every other endpoint in the controller:

```ts
@Post(':id/leave')
leaveGame(@Param('id') gameId: string, @Req() req: AuthRequest,) {
```

It should follow the same pattern:

```ts
@ApiCookieAuth()
@ApiOperation({ summary: 'Cancel a waiting game' })
@ApiParam({ name: 'id', description: 'Game ID' })
@ApiResponse({ status: 200, description: 'Game cancelled' })
@ApiResponse({ status: 400, description: 'Cannot cancel a game in progress' })
@ApiResponse({ status: 404, description: 'Game not found' })
@Post(':id/leave')
```

backend/src/modules/game/game.service.ts
```diff
@@ -293,6 +295,10 @@ export class GameService {
     const game = this.getMutableGameById(gameId);
     const role = getPlayerRoleByUserId(game, userId);
 
+    // playing
+    if (game.status === 'playing')
+      throw new BadRequestException('Cant delete game pplaying');
```
Mikan95, line 305: ### Typo in error message (`game.service.ts`)

**File:** `backend/src/modules/game/game.service.ts`

```ts
throw new BadRequestException('Cant delete game pplaying');
```

> `pplaying` has a double p, and `Cant` is missing an apostrophe.  

**Suggested message:** 
```text
Cannot cancel a game that is already in progress
```

frontend/src/pages/Chat.tsx
```diff
@@ -0,0 +1,43 @@
...
+                                onClick={() => navigate("/")}>
+                                    {t("buttons.backHome")}
+                            </Button>
+                        </>
+                    }
+                />
+            </section>
+        )
+    }
+}
```
Mikan95, line 43: ### `Chat.tsx` returns nothing for logged-in users

The function handles the `!user` case with a return, but has no return statement for when the user is logged in. The component renders nothing. 

The chat route is currently commented out in `App.tsx` so this does not affect users today, but the file will cause a blank page if the route is ever re-enabled as-is.

---

I know we discussed during our weekly meeting that we may remove the `chat` page entirely and just keep the chat as a sidebar. If that is the case it may be better to simply delete this file? Or is there another reason this file exists?

simo1616 (approved):
Thanks for the fixes!

I checked the points from my previous review:
- Spanish translations are now working correctly on the pages I tested
- cancel game no longer clears the lobby UI before the backend confirms the request
- the debug `console.log` was removed
- avatar upload texts are now translated
- the profile save toast now uses a success style

The Play / Spectate flow and the UI changes still work well after these fixes.

Looks good to me, approved

frontend/src/components/ui/GameRules.tsx
```diff
@@ -41,11 +47,19 @@ export function GameRules() {
 function Rule({ icon, title, desc }: any) {
```
Mikan95, line 47: ### `Rule` component uses `any` type (`GameRules.tsx`)

Using `any` disables TypeScript checks for this component. Every other component in `components/ui/` uses a proper type — this one is the only exception.

The fix is to declare an interface just above the `Rule` function, at the bottom of the same file (since `Rule` is a private helper used only inside `GameRules.tsx`, the interface doesn't need to go anywhere else):

```ts
interface RuleProps {
    icon: React.ReactNode;
    title: string;
    desc: string;
}

function Rule({ icon, title, desc }: RuleProps) {
```

`React.ReactNode` is the right type for `icon` — it covers any JSX element like `<User size={18} />`.

Mikan95 (commented):
## Review Summary

Great PR overall. The lobby system, live games page and navigation rework are all well structured. There is one critical bug that breaks Spanish translations and a few smaller things worth fixing. Details below.

| # | File | Line | Summary |
| --- | --- | --- | --- |
| 1 | `frontend/src/i18n/es.json` | 103 | Missing `}` closes `play` block — all Spanish translations from `profile` onward are broken |
| 2 | `frontend/src/pages/Play.tsx` | 79 | Polling interval starts regardless of auth state — fires 401 requests every 1.5s when not logged in |
| 3 | `frontend/src/pages/Play.tsx` | 64 |  `handleCancelGame` doesn't check `response.ok` — UI clears game state even if the server call failed |
| 4 | `backend/src/modules/game/game.controller.ts` | 57 |  `POST :id/leave` has no Swagger decorators — invisible in API docs |
| 5 | `backend/src/modules/game/game.service.ts` | 300 |  Typo in error message: `'Cant delete game pplaying'` |
| 6 | `frontend/src/pages/Chat.tsx` | 43 |  New file — no return for authenticated users; consider removing if chat page is no longer planned |
| 7 *(pre-existing)* | `frontend/src/components/ui/GameRules.tsx` | 47 |  `Rule` props typed as `any` — only exception in the whole `ui/` directory |
| 8 *(pre-existing)* | Multiple files | — | **see comment below** |

---

### Issue 8 - `User` interface is duplicated across many files

`LiveGames.tsx`, `Play.tsx`, `Sidebar.tsx`, `Chat.tsx` and others each define their own local `User` interface with slightly different optional fields. The root cause seems to be `Dashboard.tsx`, which stores the user as `useState<any>(null)` — because the context is untyped, each component likely ended up defining its own local shape instead of sharing one.

One option to clean this up would be to add a shared narrow type to `user.types.ts` specifically for the outlet context — something like:

```ts
// Non-sensitive fields only — for passing through outlet context
export type UserDisplay = {
    username: string;
    avatar: string;
    bio: string | null;
    wins: number;
    losses: number;
    draws: number;
    xp: number;
};
```

Then `Dashboard.tsx` could use `useState<UserDisplay | null>(null)` and all components could import `UserDisplay` from `user.types.ts` rather than each defining their own copy.

I'd suggest not using the already existing `User` type from `user.types.ts` for the context, since it includes sensitive fields such as `email` and `isTwoFactorEnabled`.

## Extra minor issues not listed in the table above

### 'Broken' buttons

These may be intentional at this state or you may intend to make these changes in a subsequent PR. However, I noticed the following buttons lead to the wrong places or are 'broken':

| Page | Button | Behaviour |
| --- | --- | --- |
| Home | `Join a game` | Brings user to a white page at path: `$DOMAIN:$PORT/lobby` | 
| Home | clicking on the `Play Game` Card | Same as `Join a game` above. |
| Home | `Watch a game` | Creates a game and brings user to the Game ID URL page |


### Inconsistent Icons when Logged out and browsing through pages

When the User is logged out I noticed certain pages show the wrong or inconsistent Icons in the Card elements.

- Play page shows History Icon instead of Play Icon
- Profile shows '?' (Consistent with `about` card on `Home` page so this may be intentional?)

This is definitely a minor issue, but I think we should make a decision to keep this consistent when logged out. I see two options but am open to any other ideas:
- display question marks everywhere
- display the icons that match the page (profile icon for profile/about card, history icon for history card, etc.)

Let me know what you think!

PR #317, opened 2026-05-22
[INFRA] - Make `nuke` idempotent

Mikan95:
As requested by @simo1616, delivered with extra love 😂

## Checklist

- [x] Target branch is `dev`

### **INFRA:**

- `make/clean.mk`: added `2>/dev/null || true` to both `docker compose down` commands in `_nuke-apply`, and a warning message when `.env` files are not found

## Why

`make nuke` deletes `.env.dev` and `.env.prod` at the end of its run. A second `nuke`
would then fail immediately — `docker compose` requires those files via `--env-file`
and exits with an error if they are missing, which Make treats as a build failure and halts.

`2>/dev/null` suppresses the error message from `docker compose`. `|| true` replaces
the non-zero exit code with success — Make stops execution on any non-zero exit, so
this is the idiomatic one-liner for "try this, but do not stop if it fails." The
volume, image, and cache pruning that follows does not need the env files and now
always runs regardless.

## Testing

- [x] Docker build successful

### Test 1 — nuke is idempotent

**1. Run nuke once to wipe everything:**

```bash
make nuke
```

**2. Run nuke a second time and confirm it completes without errors:**

```bash
make nuke
```

> Expected: warning message about missing `.env` files, followed by Docker pruning output. No Make error.

---

### DIFF
--- a/make/clean.mk
+++ b/make/clean.mk
@@ -37,7 +37,7 @@ nuke: ## Full wipe — stops stack, removes volumes + images, deletes .env + sec
 		*) \
 			exit 0 ;; \
 	esac
-	
+
 _nuke-apply: # Runs a full wipe
   	# ── Prompt for Postgres Image Removal ────────────────────────────────────
 	@printf "$(CYAN)Remove postgres:$(POSTGRES_VERSION)? Skip if rebuilding soon$(RES) [y/N] "; read ans; \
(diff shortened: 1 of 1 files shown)


simo1616 (approved):
Approved ✅

Tested it twice and everything works fine.

The issue with running `make nuke` a second time is fixed now. Even if the `.env` files are already gone, the script does not crash anymore and Docker cleanup still runs correctly.

Thanks for taking care of this extra task, and I hope it did not take you too much time 🙏

Small fix, but useful ^^ 

thanks bro

PR #325, opened 2026-05-25
[BUG] - Auto-login user on registration

Mikan95:
## Checklist

- [x] Target branch is `dev`

### **Backend:**

- Updated `register()` in `auth.controller.ts` to be `async` and accept `@Res({ passthrough: true })`
- After successful registration, calls `signTokenForUser()` and sets the JWT as an `access_token` cookie with the same options used by the login endpoint

## Why

After registering, the user was not logged in. The frontend received a success response but no JWT cookie was set, so the subsequent call to `/auth/me` returned 401 and the user was redirected back to the login page.

Fixing this in the controller keeps the change minimal. Registration and login now share the same cookie-setting path without duplicating business logic in the service.

## Testing

- [x] Docker build successful

### Prod Test 1 — Registration sets cookie and auto-logs in user

**1. Start the stack:**

```bash
make prod
```

**2. Open browser DevTools (Network tab) before registering.**

**3. Register a new account via the UI.**

> **Note:** Confirm no 401 errors appear in the console and that you land on the home/dashboard page without being redirected back to login.

---

### DIFF
--- a/backend/src/auth/auth.controller.ts
+++ b/backend/src/auth/auth.controller.ts
@@ -35,30 +35,22 @@ export class AuthController {
 	) {}
 
 	@Public()
-	@ApiOperation({ summary: 'Register a new user' })
-	@ApiBody({ type: RegisterDto })
-	@ApiResponse({
-		status: 201,
-		description: 'User registered successfully',
-	})
-	@ApiResponse({
-		status: 400,
-		description: 'Invalid input data',
-	})
 	@Post('register')
+	@ApiOperation({ summary: 'Register a new user and start a session' })
+	@ApiBody({ type: RegisterDto })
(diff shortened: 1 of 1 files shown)


Nico-Ry (approved):
Tested after resolving the merge conflicts and updating the register endpoint to reuse the current auth cookie architecture.

Everything works correctly now:

* registration automatically creates a valid authenticated session
* `access_token` cookie is properly set after register
* user stays logged in after account creation
* `/auth/session` and `/auth/me` work correctly after registration
* no unexpected 401 redirect loop anymore
* existing login/logout/2FA flow still behaves correctly

I also updated  with the comit `a13513e` this implementation , now it reuses the centralized cookie configuration already present in `dev` instead of duplicating cookie settings inline.

Looks good to merge 👍

PR #332, opened 2026-05-27
[BACKEND] Fix Prisma seed error when deleting dummy games

simo1616:
## Summary

This PR fixes a Prisma/PostgreSQL seed error that happened when starting the backend in prod.

The app was still starting after the error, but the seed was not clean and it could look bad during evaluation.

## Problem

The seed was trying to delete existing `Game` rows between dummy users.

But some games already had related `Move` rows in the database.

Because of the foreign key between `Move` and `Game`, PostgreSQL blocked the delete with this error:

```txt
update or delete on table "Game" violates RESTRICT setting of foreign key constraint "Move_gameId_fkey" on table "Move"
````

## What we changed

We fixed the seed deletion order.

Before deleting dummy games, the seed now deletes the related moves first:

```ts
await prisma.move.deleteMany(...);
await prisma.game.deleteMany(...);
```

We also use the IDs returned by `upsert` for dummy users instead of relying only on usernames.

This makes the seed safer if a dummy user was renamed manually.

## Why this solution

We chose the safer fix for now.

Instead of changing `schema.prisma` and adding cascade delete globally, we only fixed the seed logic.

This avoids changing the global behavior of game deletion in the app just before evaluation.

## Main changes

* delete related `Move` rows before deleting dummy `Game` rows
* keep the Prisma schema unchanged
* use dummy user IDs returned by `upsert`
* make the seed cleaner and more stable
* remove the Prisma foreign key error from prod logs

## Result

The backend prod startup is now clean:

* seed runs successfully
* dummy users are created/updated
* old dummy games and moves are deleted correctly
* new dummy games are added
* user stats and XP are updated
* backend starts normally after the seed

## How to test

* Run `make prod`
* Check backend logs with:

```bash
make p-logs-back
```

* Confirm the logs show:

```txt
>>> Seeding successful!
Nest application successfully started
```

* Confirm there is no more:

```txt
DriverAdapterError
violates RESTRICT
Move_gameId_fkey
```

* Login with a dummy account, for example:

```txt
dummy0@mail.com
Dummy123
```

* Check that history / leaderboard still work
* Restart prod a second time to confirm the seed stays clean on repeated starts

## Notes

This PR only fixes the seed behavior.

It does not add `onDelete: Cascade` on the Prisma relation, so normal game deletion behavior in the app is not changed. This should help keep prod logs clean for evaluation without touching the global Prisma relation behavior.

### DIFF
--- a/backend/prisma/seed.ts
+++ b/backend/prisma/seed.ts
@@ -1,9 +1,8 @@
-import { Pool } from "pg";
-import { PrismaPg } from "@prisma/adapter-pg";
+import { Pool } from 'pg';
+import { PrismaPg } from '@prisma/adapter-pg';
 import { PrismaClient } from '@prisma/client';
 import * as bcrypt from 'bcrypt';
 import * as fs from 'node:fs';
-
 
 // ── Define Prisma Client ────────────────────────────────────────────────────
 const backendPw = fs.readFileSync('/run/secrets/backend_pw', 'utf8').trim();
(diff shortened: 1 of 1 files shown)


jdecarro (approved):
## Review

Validated successfully:

- Prisma seed runs cleanly during prod startup
- dummy users are properly created/updated
- old dummy games are removed correctly
- related moves are deleted before game cleanup
- repeated prod restarts no longer generate foreign key errors
- leaderboard and match history still behave correctly after reseeding
- backend starts normally after seed execution

The fix is clean and well scoped.

I especially agree with the decision to avoid adding global `onDelete: Cascade` behavior in Prisma right before evaluation. Keeping the schema unchanged while fixing the seed deletion order is the safest approach here and avoids unintended side effects on normal game deletion behavior.

Using the IDs returned by `upsert` instead of relying only on usernames also makes the seed logic more robust and future-proof.

The backend logs are now clean in production and no longer show:

- `DriverAdapterError`
- `violates RESTRICT`
- `Move_gameId_fkey`

Very solid stabilization/fix !

PR #333, opened 2026-05-28
[FIX]: cleanup waiting game when creator disconnects

simo1616:
## Summary

Fix waiting games staying active after the creator disconnects.

## Changes

- Added cleanup for waiting games owned by a disconnected user.
- When a user has no more presence sockets, we delete their waiting game if:
  - the game is still `waiting`
  - the user is the creator/player X

## Why

A waiting game should not stay visible if the creator fully left the app.

## Behavior

- Another tab still open → game is kept.
- Game already `playing` → game is kept.
- Only the waiting game owner can trigger this cleanup.

## Tradeoff

Simple fix: if the creator refreshes and presence drops to zero briefly, the waiting game can be deleted.  
A delayed cleanup would avoid that, but adds more complexity.

## Test

- Create a game, close/logout creator → waiting game disappears.
- Create a game, keep another tab open → game stays.
- Start a game, disconnect a player → playing game stays.

### DIFF
--- a/backend/src/modules/game/game.service.ts
+++ b/backend/src/modules/game/game.service.ts
@@ -335,6 +335,17 @@ export class GameService {
 		return this.activeGame.delete(gameId);
 	}
 
+	deleteWaitingGameByOwner(userId: number): void {
+		const active = this.findActiveGameByUserId(userId);
+		if (!active) return;
+
+		const [gameId, game] = active;
+		if (game.status === 'waiting' && game.players.X.ownerUserId === userId) {
+			this.activeGame.delete(gameId);
+			this.presenceService.emitActiveGameUpdated(userId, null);
+		}
+	}
+
 	/**
 	 * Persists a completed game to the database.
 	 * Maps the in-memory game state (X/O players) to the MatchesService format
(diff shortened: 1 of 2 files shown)


nasdhn (approved):
## 🔍 PR Review

I tested this PR and the behavior is working correctly.

The cleanup logic for waiting games after a full user disconnect behaves as expected and fixes the issue cleanly without impacting active games.

### ✅ What was validated

* Creating a waiting game then fully disconnecting removes the waiting game correctly
* Keeping another tab open preserves the waiting game
* Playing games are not deleted on disconnect
* Cleanup only affects the waiting game owner
* Presence behavior remains stable
* No visible regression during game flow tests

### 🧠 Notes

The implementation is simple and coherent with the current architecture.

The tradeoff mentioned in the PR description also makes sense:
a very short disconnect during refresh can remove the waiting game, but the current solution avoids introducing unnecessary complexity for now.

Good balance between:

* simplicity
* readability
* expected behavior

### 🚀 Conclusion

Everything is functioning correctly from my tests.
The fix resolves the stale waiting game issue cleanly and keeps the realtime behavior consistent.

